PR #6, opened 2020-01-14
kubeconfig, disk_size and autoscaler

omerfsen:
Although there are a few more fix required such as

1) Private and Public subnet should be labeled with "kubernetes.io/role/internal" = "1"  "kubernetes.io/role/internal-elb" = "1"  and  accordingly

2) Autoscaler role should be added and attached to worker nodes

3) Ingress Role should be added accordingly so Ingress Controller works

4) EBSCsiDriverIAMPolicy should be added for the future

5) "CloudWatchAgentServerPolicy" should be added for Cloudwatch Container Insights  (Not required but nice to have)

6) External-DNS' required policy can be added 

7) Optional cluster-autoscaler deployment can be made so Node Group AutoScaling can work by default without additional steps

### DIFF
--- a/examples/complete/fixtures.us-east-2.tfvars
+++ b/examples/complete/fixtures.us-east-2.tfvars
@@ -20,6 +20,6 @@ min_size = 2
 
 disk_size = 20
 
-kubeconfig_path = "${HOME}/.kube/config"
+kubeconfig_path = "/.kube/config"
 
 kubernetes_labels = {}
--- a/examples/complete/fixtures.us-east-2.tfvars
+++ b/examples/complete/fixtures.us-east-2.tfvars
@@ -20,6 +20,6 @@ min_size = 2
 
 disk_size = 20
 
-kubeconfig_path = "/.kube/config"
+kubeconfig_path = "${HOME}/.kube/config"
 
 kubernetes_labels = {}
--- a/main.tf
+++ b/main.tf
@@ -3,6 +3,12 @@ locals {
     var.tags,
     {
       "kubernetes.io/cluster/${var.cluster_name}" = "owned"
+    },
+    {
+      "k8s.io/cluster-autoscaler/${var.cluster_name}" = "owned"
+    },
+    {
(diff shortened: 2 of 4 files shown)


examples/complete/fixtures.us-east-2.tfvars
```diff
@@ -20,6 +20,6 @@ min_size = 2
 
 disk_size = 20
 
-kubeconfig_path = "/.kube/config"
+kubeconfig_path = "${HOME}/.kube/config"
```
aknysh: these variables are for the automated test.
`/.kube/config` is ok since we are running it on CodeFresh container

main.tf
```diff
@@ -3,6 +3,12 @@ locals {
     var.tags,
     {
       "kubernetes.io/cluster/${var.cluster_name}" = "owned"
+    },
+    {
```
aknysh: @omerfsen can you provide references to documentation where you got these annotations from?
thanks

omerfsen: https://docs.aws.amazon.com/eks/latest/userguide/cluster-autoscaler.html I have even setup a test cluster with your terraform. Even if you install cluster-autoscaler without these tags autoscaling does not work

omerfsen: https://docs.aws.amazon.com/eks/latest/userguide/cluster-autoscaler.html

examples/complete/fixtures.us-east-2.tfvars
omerfsen: I do run it on my Jenkins. So it would be nice to add ${HOME} so even on container it will use / as the home dir of root user on Docker or will use /var/lib/jenkins on Jenkins so it will be multiple CI/CD compliant if you add ${HOME}

aknysh: @omerfsen thanks
Note that this file is our example/test (variables for the test) which we deploy automatically on AS using Terratest, so it's not related to the module and how you run it.
Please don't change it in this PR so we don't mix different things in one PR.

omerfsen:
All changes you have requested are done

PR #14, opened 2020-04-01
Allow node group to assume roles

maxgio92:
## what
Allow to add IAM permissions to the nodes to assume every roles.

## why
In order to make the pods the first-class citizens and authenticate them in AWS using IAM roles (i.e. short-time-lived tokens) the nodes need to assume every role. 

## references
For instance, the classic kube2iam needs it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,7 @@ Available targets:
 | desired_size | Desired number of worker nodes | number | - | yes |
 | disk_size | Disk size in GiB for worker nodes. Defaults to 20. Terraform will only perform drift detection if a configuration value is provided | number | `20` | no |
 | ec2_ssh_key | SSH key name that should be used to access the worker nodes | string | `null` | no |
+| enable_assume_all_roles | Whether to enable node group to assume every IAM role | bool | `false` | no |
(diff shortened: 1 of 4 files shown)


Nuru:
@maxgio92 You should not be using kube2iam with EKS, you should be using [IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html). Even if you are using kube2iam, you should be using IAM Roles for Service Accounts to give kube2iam access to the roles it needs to assume rather than granting all the instances access to all the roles, which is effectively turning off all security for the cluster.

maxgio92:
@Nuru I agree. I know that EKS provides the OIDC endpoint and that at the same time this is not a good security practice.

At the same time sometimes it isn’t possible to use IAM roles for service accounts in an easy way (for example when you can’t customise the Service account resources in order to add the proper annotations for EKS/IAM).

BTW thank you, honestly I didn’t think about use them only for the kube2iam daemonset :)
So, I perfectly agree on closing this PR, this isn’t a good solution.

Issue #24, opened 2020-08-26
EKS Node Launch Template support


dotCipher:
## Describe the Feature

The [eks-node-group](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group#launch_template) resource now supports launch templates.  

See associated [aws post](https://aws.amazon.com/blogs/containers/introducing-launch-template-and-custom-ami-support-in-amazon-eks-managed-node-groups/) as well as [eks launch templates docs](https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html) for more information.

## Expected Behavior

1. Define `launch_template` somewhere in call-site of this module
1. Reference `launch_template` output within attribute of `cloudposse/terraform-aws-eks-node-group`

## Use Case

I'd like to be able to add user data (ie. install SSM agent) to my EKS managed node groups, and I am unable to do that without adding launch templates.  

## Describe Ideal Solution

There should be a `launch_template` attribute that is 'settable' for the module

## Alternatives Considered

Only alternative would be to remove dependency to this module and hand write it 😛 

## Additional Context

- [AWS discussion of adding launch template support to EKS](https://github.com/aws/containers-roadmap/issues/585)
- [`launch_template` terraform resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template)

dotCipher:
Seems like this might be dependant on the version bump PR: https://github.com/cloudposse/terraform-aws-eks-node-group/pull/25

dotCipher:
FWIW, I'd gladly help write the PR, can someone point me in the right direction for the testing workflow used? ie. what commands to run, etc?

danjbh:
I'm going to push to get this knocked out today and would appreciate another set of eyes & testing when the new PR ready. I'll post it here once it's ready. Stay tuned!

danjbh:
Hey there @dotCipher , I have a first draft of this prepped that allows you to...

1. Completely override the launch template by passing `aws_launch_template.foo.id` and `launch_template.foo.version` to the module
2. Override *only* the userdata by passing it to the module as a typical, b64 encoded string
3. Not do anything extra and we'll use a minimal, default `aws_launch_template` (and default userdata) -- the primary enhancement here being that it adds the `tag_specifications` config option that will pass on our tags to `instances`, `volumes`, and `elastic-gpu`

[...]

PR #25, opened 2020-08-27
Bumping AWS provider version to ~> 3.0.0

danjbh:
## what
* Bumping our AWS provider to version 3

## why
* Allows us to take advantage of some newer Terraform features when deploying EKS
* For example, this will provide us with the ability to specify custom launch templates in `aws_eks_node_group`

#### https://github.com/aws/containers-roadmap/issues/585

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ Available targets:
 | Name | Version |
 |------|---------|
 | terraform | >= 0.12.0, < 0.14.0 |
-| aws | ~> 2.0 |
+| aws | ~> 3.0 |
 | local | ~> 1.3 |
 | template | ~> 2.0 |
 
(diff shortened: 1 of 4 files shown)


danjbh:
I've got some more work to do on this one -- stay tuned!

dotCipher:
Excited to have seen this was already being worked on! I actually just got a use-case for using launch templates and submitted the issue: #24

danjbh:
I was originally thinking that we could start by bumping the AWS versions for any modules commonly used in this pattern (e.g. `terraform-aws-eks-cluster`), but instead I'm going to create a new branch and start working on the launch template functionality today. I'll update #24 with the new PR when it's ready (hopefully later today) 👍

PR #27, opened 2020-08-28
Adding support for launch templates & userdata parameters

danjbh:
## what
* Adding default launch template configuration
* Adding ability to provide your own launch template by overriding the launch template id & version
* Adding dynamic config options for user_data
* Bumping various upstream module versions & tests
* Keeping `instance_types` as a list but adding TF 0.13 variable validation

## why
In previous version of the AWS provider (2.x), you could not define your own launch template for `aws_eks_node_group`. Additionally, the tags specified in the `aws_eks_node_group` definition were not being passed down to the EC2 instances created by the ASG, which made tasks like monitoring and cost tracking difficult.

The latest versions of the AWS provider (3.x) give us the ability to specify our own launch template directly from `aws_eks_node_group`, which allows us to set our own options (e.g. `tag_specifications`. `user_data`, etc.).

This also should satisfy the requests in #24 

## references
* https://github.com/aws/containers-roadmap/issues/585
* https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh


README.md
```diff
@@ -146,7 +146,7 @@ For automated tests of the complete example using [bats](https://github.com/bats
       attributes                = var.attributes
       tags                      = var.tags
       subnet_ids                = module.subnets.public_subnet_ids
-      instance_types            = var.instance_types
+      instance_type             = var.instance_type
```
danjbh, line 149: So I really struggled with this one..

I don't like the impact this change has to existing folks, but on the other hand I also don't like misleading the user into thinking that we can accept multiple instance types when it's unsupported. 

The gist of it is, now that we're defining our own launch template, we have to specify the instance type in `aws_launch_template` instead of `aws_eks_node_group`. And well, the `aws_launch_template` resource accepts a string (i.e. only one instance type). 

In previous versions, we were passing the `instance_types` into `aws_eks_node_group` as a list, although the documentation states that the EKS API "only accepts a single value in the set."

Anyhow, it sounds like multiple instance types may be supported at some point in the future, but for now it seems prudent to limit the user to a single type. I'm open to further input if anyone has any other ideas -- it'd be easy enough to revert/change if needed anyhow.

README.md
```diff
@@ -172,21 +172,20 @@ Available targets:
...
 ```
 <!-- markdownlint-restore -->
-<!-- markdownlint-disable -->
 ## Requirements
 
 | Name | Version |
 |------|---------|
 | terraform | >= 0.12.0, < 0.14.0 |
-| aws | >= 2.0, < 4.0 |
+| aws | ~> 3.0 |
```
danjbh, line 180: We need 3.0 for launch template support.

main.tf
```diff
@@ -1,8 +1,8 @@
 locals {
   enabled = module.this.enabled
 
-  tags = merge(
-    var.tags,
+  node_group_tags = merge(
```
danjbh, line 4: I modified this because the cluster-level tags were being applied to all resources in the module (e.g. IAM roles, policies, etc.), and we probably don't want them there.

README.md
dotCipher: Yup, totally noticed that multiple instance types weren't supported when I tried providing multiple originally here

README.md
```diff
@@ -208,10 +208,13 @@ Available targets:
...
+| instance\_type | Instance type associated with the EKS Node Group. Defaults to "t3.medium". Terraform will only perform drift detection if a configuration value is provided | `string` | n/a | yes |
 | kubernetes\_labels | Key-value mapping of Kubernetes labels. Only labels that are applied with the EKS API are managed by this argument. Other Kubernetes labels applied to the EKS Node Group will not be managed | `map(string)` | `{}` | no |
 | kubernetes\_version | Kubernetes version. Defaults to EKS Cluster Kubernetes version. Terraform will only perform drift detection if a configuration value is provided | `string` | `null` | no |
 | label\_order | The naming order of the id output and Name tag.<br>Defaults to ["namespace", "environment", "stage", "name", "attributes"].<br>You can omit any of the 5 elements, but at least one must be present. | `list(string)` | `null` | no |
+| launch\_template\_id | The ID of a custom launch template to use for the EKS node group. | `string` | `null` | no |
```
dotCipher, line 215: We can also provide a `launch_template.name` based on the docs, do we only want to allow `id` instead?

main.tf
```diff
@@ -113,27 +119,56 @@ resource "aws_iam_role_policy_attachment" "existing_policies_for_eks_workers_rol
...
+  dynamic "tag_specifications" {
+    for_each = ["instance", "volume", "elastic-gpu"]
+    content {
+      resource_type = tag_specifications.value
+      tags          = local.node_group_tags
+    }
+  }
+
+  # Override the default userdata if input is provided
+  # If none is provided, this value defaults to 'null', which defaults to the typical EKS node group userdata
```
dotCipher, line 144: Will this clobber the default EKS node group userdata if provided? Is there anything important that runs via the default userdata or is it just always empty?

danjbh: The default userdata runs `bootstrap.sh` w/ the pertinent options to attach the node(s) to the EKS cluster. If you're using a custom launch template, or just a userdata override, you would have to include the entire script basically. 

And actually, I just realized that we have a solution in `terraform-aws-eks-workers` that will allow us to parameterize it, so let me add that and refactor this a little bit. Stay tuned!

README.md
danjbh: Yeah, I saw that, although the `id` seemed definitive enough to be honest. I don't know if I like the idea of adding the template just by the name, but I'll chew on this a bit.

danjbh:
I still need to port over the logic from `terraform-aws-eks-workers` to parameterize userdata, so I'm going to move this to draft.

README.md
dotCipher: Honestly the opinionated view of only allowing id doesn't seem all that bad at all, it's definitely intuitive

main.tf
dotCipher: Sounds great

danjbh: All done if you want to have a look!

dotCipher:
This will require TF 0.13 right?

danjbh:
> This will require TF 0.13 right?

Yep yep.. we had a discussion about it and the preference was to go w/ 0.13 and use the new validation functionality basically.

Gowiem:
This is looking awesome @danjbh! Went down the rabbit hole of trying to figure why my node-group instances didn't have `name` tags and stumbled across this PR, so I'm stoked for this to land. If my project was already on 0.13 I'd give it a test run and then give you a review, but without being able to test this many changes I'll hold off for a bit until I can give it a real pass.

userdata.tpl
```diff
@@ -0,0 +1,21 @@
...
+# userdata for EKS worker nodes to properly configure Kubernetes applications on EC2 instances
+# https://docs.aws.amazon.com/eks/latest/userguide/launch-workers.html
+# https://aws.amazon.com/blogs/opensource/improvements-eks-worker-node-provisioning/
+# https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh#L97
+
+${before_cluster_joining_userdata}
+
+export KUBELET_EXTRA_ARGS="${kubelet_extra_args}"
+
+/etc/eks/bootstrap.sh --apiserver-endpoint '${cluster_endpoint}' ${bootstrap_extra_args} '${cluster_name}'
```
Nuru, line 17: `--apiserver-endpoint` is not needed and documented as "not valid" without also specifying cluster CA, which we also do not need/want to do.

```suggestion
/etc/eks/bootstrap.sh ${bootstrap_extra_args} '${cluster_name}'
```

main.tf
```diff
@@ -14,6 +14,22 @@ locals {
     }
   )
   aws_policy_prefix = format("arn:%s:iam::aws:policy", join("", data.aws_partition.current.*.partition))
+
+  userdata_vars = {
+    cluster_endpoint                = var.cluster_endpoint
```
Nuru, line 19: `cluster_endpoint` is not needed
```suggestion
```

main.tf
```diff
@@ -24,12 +40,11 @@ module "label" {
   # ...name-workers-blue instead of ...name-blue-workers), historically we forced "workers"
   # to the end of the attribute list, so we do it again here to maintain compatibility.
   attributes = compact(concat(module.this.attributes, ["workers"]))
-  tags       = local.tags
+  tags       = var.tags
```
Nuru, line 43: `tags = local.tags` was needed to add the extra tags. `tags = var.tags` is not needed, it is handled by `context`.
```suggestion
```

variables.tf
```diff
@@ -9,6 +9,11 @@ variable "cluster_name" {
   description = "The name of the EKS cluster"
 }
 
+variable "cluster_endpoint" {
+  type        = string
+  description = "EKS cluster endpoint"
+}
+
```
Nuru, line 16: `cluster_endpoint` is not needed
```suggestion
```

PR #29, opened 2020-08-31
Added support for custom AMIs and node taints and other kubelet args

xario:
## What
These changes are in relation to the PR about launch configurations (#27). This PR is circumventing the problem with not being able to specify the `KUBELET_EXTRA_ARGS` in user_data when not specifying `ami_id`, as noted in the PR which rolled back some of the changes (#28). 
As AWS appends a call to the `/etc/eks/bootstrap.sh` to the user_data, it is necessary to modify the bootstrap file to include the changes to the config.  This is done by  adding a hook [right before the services are started](https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh#L340), which allows for updates to the `/etc/systemd/system/kubelet.service.d/30-kubelet-extra-args.conf` file.

Here is an outline of what is happening the hook script:
* Read out the existing configuration from `30-kubelet-extra-args.conf`
* Parse the arguments in `KUBELET_EXTRA_ARGS` into an associative array
* Parse the arguments from the user supplied `KUBELET_NEW_ARGS` into the array
* Add taints from the user supplied `NEW_TAINTS` to the entry `--register-with-taints` in the array
* Add labels from the user supplied `NEW_LABELS` to the entry `--node-labels` in the array
* Combine the entries from the array into a string, and write it back to `30-kubelet-extra-args.conf`

This preserve the arguments that the `bootstrap.sh` adds to `KUBELET_EXTRA_ARGS`, while still being able to add user specified arguments.

A minor change is that I had to make to the way the security groups are added to a node group, as I got an error `Remote access configuration cannot be specified with a launch template`. They are now added to the launch configuration by using `vpc_security_group_ids`. By specifying this, it overwrites the security group AWS adds normally, that allows for communication with the control plane, which is why I had to add a data source for the `aws_eks_cluster` in order to retrieve the security group id. This change can be omitted in this PR and added as a separate PR if you like.

## Why
* This is needed in order to use managed node groups with node taints, without having to specifying an `ami_id`.

## References
* aws/containers-roadmap#585
* https://aws.amazon.com/blogs/containers/introducing-launch-template-and-custom-ami-support-in-amazon-eks-managed-node-groups/
* https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh

PS: I have not updated tests and documentation yet, as I wanted to get your reaction to the changes before continuing. 
The changes have been tested with and without specifying an `ami_id`, and with and without specifying an `ec2_ssh_key` by creating clusters using Terraform in AWS.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -15,8 +15,22 @@ locals {
   )
   aws_policy_prefix = format("arn:%s:iam::aws:policy", join("", data.aws_partition.current.*.partition))
 
+  node_labels = [
+    for item in keys(var.kubernetes_labels):
+      join("=", [item, lookup(var.kubernetes_labels, item)])
+  ]
+
   userdata_vars = {
-    before_cluster_joining_userdata = var.before_cluster_joining_userdata
+    before_cluster_joining_userdata = var.before_cluster_joining_userdata,
+    kubelet_extra_args = replace(join(" ", var.kubelet_extra_args), "'", ""),
+    node_taints = replace(join(",", var.node_taints), "'", ""),
+    node_labels = replace(join(",", local.node_labels),"'", "")
+  }
+
+  userdata_ami_vars = {
+    cluster_name = var.cluster_name,
+    node_group_name = module.label.id,
+    ami_id = var.ami_id
   }
 
   # Use a custom launch_template if one was passed as an input
@@ -133,6 +147,8 @@ resource "aws_launch_template" "default" {
     }
   }
 
+  image_id = var.ami_id
+
   instance_type = var.instance_types[0]
 
   dynamic "tag_specifications" {
@@ -143,7 +159,38 @@ resource "aws_launch_template" "default" {
     }
   }
 
-  user_data = base64encode(templatefile("${path.module}/userdata.tpl", local.userdata_vars))
+  vpc_security_group_ids = concat(
+    [data.aws_eks_cluster.eks_cluster.vpc_config[0].cluster_security_group_id],
+    var.source_security_group_ids,
+    aws_security_group.remote_access_security_group.*.id
+  )
+
+  key_name = var.ec2_ssh_key
+
+  user_data = base64encode(
+    format("%s%s",
+      templatefile("${path.module}/userdata.tpl", local.userdata_vars),
+      (var.ami_id != null ? templatefile("${path.module}/userdata_ami.tpl", local.userdata_ami_vars) : "")
+    )
+  )
+}
+
+data "aws_eks_cluster" "eks_cluster" {
+  name = var.cluster_name
+}
+
+resource "aws_security_group" "remote_access_security_group" {
+  count = var.ec2_ssh_key != null && length(var.source_security_group_ids) == 0 ? 1 : 0
+  name        = "eks-${var.cluster_name}-${module.label.id}-remote-access"
+  description = "Security group for all nodes in the nodeGroup to allow SSH access"
+  vpc_id      = data.aws_eks_cluster.eks_cluster.vpc_config[0].vpc_id
+
+  ingress {
+    from_port   = 22
+    to_port     = 22
+    protocol    = "tcp"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
 }
 
 resource "aws_eks_node_group" "default" {
@@ -152,9 +199,9 @@ resource "aws_eks_node_group" "default" {
   node_group_name = module.label.id
   node_role_arn   = join("", aws_iam_role.default.*.arn)
   subnet_ids      = var.subnet_ids
-  ami_type        = var.ami_type
+  ami_type        = var.ami_id == null ? var.ami_type : null
   labels          = var.kubernetes_labels
-  release_version = var.ami_release_version
+  release_version = var.ami_id == null ? var.kubernetes_version : null
   version         = var.kubernetes_version
 
   tags = local.node_group_tags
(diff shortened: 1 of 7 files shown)


main.tf
```diff
@@ -15,8 +15,22 @@ locals {
...
+    for item in keys(var.kubernetes_labels):
+      join("=", [item, lookup(var.kubernetes_labels, item)])
+  ]
+
   userdata_vars = {
-    before_cluster_joining_userdata = var.before_cluster_joining_userdata
+    before_cluster_joining_userdata = var.before_cluster_joining_userdata,
+    kubelet_extra_args = replace(join(" ", var.kubelet_extra_args), "'", ""),
+    node_taints = replace(join(",", var.node_taints), "'", ""),
+    node_labels = replace(join(",", local.node_labels),"'", "")
```
Nuru, line 27: I do not like the idea of blindly stripping quotes out of values. What if they have backslash escapes, for example?

main.tf
```diff
@@ -143,7 +159,38 @@ resource "aws_launch_template" "default" {
     }
   }
 
-  user_data = base64encode(templatefile("${path.module}/userdata.tpl", local.userdata_vars))
+  vpc_security_group_ids = concat(
+    [data.aws_eks_cluster.eks_cluster.vpc_config[0].cluster_security_group_id],
+    var.source_security_group_ids,
+    aws_security_group.remote_access_security_group.*.id
+  )
```
Nuru, line 166: Why is this needed?

main.tf
```diff
@@ -143,7 +159,38 @@ resource "aws_launch_template" "default" {
...
+
+  key_name = var.ec2_ssh_key
+
+  user_data = base64encode(
+    format("%s%s",
+      templatefile("${path.module}/userdata.tpl", local.userdata_vars),
+      (var.ami_id != null ? templatefile("${path.module}/userdata_ami.tpl", local.userdata_ami_vars) : "")
```
Nuru, line 173: Why is this 2 files concatenated rather than 1 file?

main.tf
```diff
@@ -143,7 +159,38 @@ resource "aws_launch_template" "default" {
...
+  description = "Security group for all nodes in the nodeGroup to allow SSH access"
+  vpc_id      = data.aws_eks_cluster.eks_cluster.vpc_config[0].vpc_id
+
+  ingress {
+    from_port   = 22
+    to_port     = 22
+    protocol    = "tcp"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
 }
```
Nuru, line 194: What is the advantage of this over the previous dynamic remote access block in `aws_eks_node_group`?

userdata.tpl
```diff
@@ -2,9 +2,134 @@ MIME-Version: 1.0
...
+}
+
+# 30-kubelet-extra-args.conf example:
+# [Service]
+# Environment='KUBELET_EXTRA_ARGS=--register-with-taints="dedicated=test:NoSchedule" --node-labels=eks.amazonaws.com/nodegroup=test-mng,other.label=other'
+
+# Get the last line and cut out the part between the single quotes
+# Example output:
+# KUBELET_EXTRA_ARGS=--register-with-taints="dedicated=test:NoSchedule" --node-labels=eks.amazonaws.com/nodegroup=test-mng,other.label=other
+ENVIRONMENT=$(tail -n1 /etc/systemd/system/kubelet.service.d/30-kubelet-extra-args.conf | cut -d\' -f2)
```
Nuru, line 40: This is brittle, in that it requires that the last line of the file be desired line and will break if/when the default changes.

userdata.tpl
```diff
@@ -2,9 +2,134 @@ MIME-Version: 1.0
...
+# Environment='KUBELET_EXTRA_ARGS=--log-file=/tmp/kubelet.log --hostname-override=dedicated-node-az1 --node-labels=eks.amazonaws.com/nodegroup=test-mng,other.label=other,new.label=new --register-with-taints=dedicated=test:NoSchedule,dedicated=new-test:NoSchedule'
+cat <<CONF > /etc/systemd/system/kubelet.service.d/30-kubelet-extra-args.conf
+[Service]
+Environment='KUBELET_EXTRA_ARGS=$${NEW_KUBELET_EXTRA_ARGS}'
+CONF
+EOF
+
+# Execute the hook script before starting the services
+sed -i "/^systemctl daemon-reload\$/i . /etc/eks/extra_args.sh" /etc/eks/bootstrap.sh
+--//
```
Nuru, line 135: While I applaud the effort to be comprehensive, this still seems to me a lot more complicated and brittle than it needs to be. I do not want to be supporting this a year from now.

userdata.tpl
```diff
@@ -2,9 +2,134 @@ MIME-Version: 1.0
 Content-Type: multipart/mixed; boundary="//"
 
 --//
-Content-Type: text/x-shellscript; charset="us-ascii"
-#!/bin/bash
+Content-Type: text/x-shellscript
+Content-Type: charset="us-ascii"
```
Nuru, line 6: AFAIK you are not allowed to have multiple `Content-Type` headers. For sure I see no advantage to it.

userdata.tpl
```diff
@@ -2,9 +2,134 @@ MIME-Version: 1.0
...
 
 ${before_cluster_joining_userdata}
 
---//--
+cat << 'EOF' > /etc/eks/extra_args.sh
+#!/usr/bin/env bash
```
Nuru, line 11: While arguably `#!/usr/bin/env bash` is more portable than `#!/bin/bash` it is less secure and if the AMI you are using does not have `bash` installed as `/bin/bash` then what are the chances that is it also using `/etc/systemd/system/kubelet.service.d/30-kubelet-extra-args.conf`? Not great. So I would leave this as `#!/bin/bash`

Nuru:
@xario Thank you for your contribution. We already have in the works something similar, but more along the lines of our other modules. I think your version is more complex and sophisticated than we need or want to support long term, so I am going to close this PR. Expect support for labels and taints in less than a week from now.

PR #30, opened 2020-09-03
Fix for remote access and lifecycle issues 

woz5999:
## what

- Create remote access security group for launch template when necessary to enable SSH access
- Resolve some lifecycle dependency issues caused by create-before-destroy

## why

- Solves the issue where `remote_access` is not valid for node groups that specify launch templates. If you specify an ssh key or source security group ids with the current state of the module, it will throw an error and prevent node group creation.
- Fixes some situations where Terraform would fail to apply a plan due to a dependency cycle


woz5999:
@Nuru @danjbh This PR has fixes some module issues that make it unusable for users trying to specify an ssh key and additional remote access security groups and also fixes a regression that breaks govcloud support.

main.tf
```diff
@@ -25,6 +25,20 @@ locals {
...
   }
+
+  eks_node_group_resources = merge(aws_eks_node_group.default.*.resources,
+    {
+      "remote_access_security_group_id" = aws_security_group.remote_access.id
+    }
+  )
+
+  security_group_ids = concat(
+    list(data.aws_eks_cluster.eks_cluster.vpc_config[0].cluster_security_group_id),
```
woz5999, line 36: needed to enable communication with the EKS control plane

main.tf
```diff
@@ -25,6 +25,20 @@ locals {
     id             = coalesce(var.launch_template_id, aws_launch_template.default[0].id)
     latest_version = coalesce(var.launch_template_version, aws_launch_template.default[0].latest_version)
   }
+
+  eks_node_group_resources = merge(aws_eks_node_group.default.*.resources,
```
woz5999, line 29: this is some formatting to maintain backward compatibility with outputs.

main.tf
```diff
@@ -57,6 +71,14 @@ data "aws_iam_policy_document" "assume_role" {
   }
 }
 
+data "aws_subnet" "default" {
+  id = var.subnet_ids[0]
```
woz5999, line 75: this is to prevent introducing an additional required variable. happy to modify if that's a more preferred route.

main.tf
```diff
@@ -170,14 +194,6 @@ resource "aws_eks_node_group" "default" {
     version = local.launch_template.latest_version
   }
 
-  dynamic "remote_access" {
```
woz5999, line 173: this block is not supported when using launch templates. currently breaks the entire module if `ec2_ssh_key` or `source_security_group_ids` is specified

main.tf
```diff
@@ -195,3 +211,34 @@ resource "aws_eks_node_group" "default" {
...
+resource "aws_security_group_rule" "public_ssh" {
+  count             = local.enabled && var.ec2_ssh_key != null && length(var.source_security_group_ids) < 1 ? 1 : 0
+  from_port         = 22
+  protocol          = "tcp"
+  security_group_id = aws_security_group.remote_access[0].id
+  to_port           = 22
+  type              = "ingress"
+
+  cidr_blocks = [
+    "0.0.0.0/0"
```
woz5999, line 232: this emulates the default behavior of non-launch template node groups that specify a `remote_access` ssh key but no source security group.

main.tf
```diff
@@ -25,6 +25,22 @@ locals {
...
+        remote_access_security_group_id = join("", aws_security_group.remote_access.*.id)
+      })
+    ]
+  ]
+
+  security_group_ids = concat(
+    list(data.aws_eks_cluster.eks_cluster.vpc_config[0].cluster_security_group_id),
+    var.ec2_ssh_key != null ? aws_security_group.remote_access.*.id : []
+  )
+
```
Nuru, line 41: I am still figuring out the best way to allow variables to be empty but easy to test. If you have a better idea, I would love to hear it; for now what I have is:
```suggestion
    local.ec2_ssh_key != null ? aws_security_group.remote_access.*.id : []
  )

  ec2_ssh_key = var.ec2_ssh_key == "" ? : null ? var.ec2_ssh_key
```

main.tf
```diff
@@ -133,10 +157,12 @@ resource "aws_launch_template" "default" {
     }
   }
 
-  instance_type = var.instance_types[0]
+  instance_type          = var.instance_types[0]
+  key_name               = var.ec2_ssh_key
+  vpc_security_group_ids = local.security_group_ids
```
Nuru, line 162: In case we get the security group list wrong, I do not want to specify it if we do not have to.
```suggestion
  key_name               = local.ec2_ssh_key
  vpc_security_group_ids = local.ec2_ssh_key != null ?  local.security_group_ids : null
```

main.tf
```diff
@@ -133,10 +157,12 @@ resource "aws_launch_template" "default" {
...
 
   dynamic "tag_specifications" {
-    for_each = ["instance", "volume", "elastic-gpu"]
+    for_each = local.tag_spec
```
Nuru, line 165: This is handled in v0.11.0

Nuru (changes requested):
Thank you for this submission. 

We added a lot in v0.11.0. In particular, it should solve your tag problem. We did not address SSH access when creating a launch template because it seems like there are a lot of ways for that to go wrong and we did not have a use case or a test for it.  If you do not care about propagating tags to instances and EBS volumes, then you do not need a launch template and v0.11.0 should do what you need. Otherwise, please review 0.11 and the comments here, rebase and revise the PR, and point us to any documentation you can to assure us that the security group configuration for the launch template is correct for all the various use cases.

main.tf
```diff
@@ -25,6 +25,22 @@ locals {
     id             = coalesce(var.launch_template_id, aws_launch_template.default[0].id)
     latest_version = coalesce(var.launch_template_version, aws_launch_template.default[0].latest_version)
   }
+
+  eks_node_group_resources = [for resource in aws_eks_node_group.default.*.resources :
```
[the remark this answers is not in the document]
woz5999: the sg id is needed as an output for users that might want to add rules to it outside the module. i was appending to resources to match prior behavior. happy to put it as its own output. 
it's important to have as an output for users that might want to add additional rules outside the module.  if you're ok breaking backward compatibility, this can all be updated and users can just provide an sg id to the module to attach to the launch template and bypass most of the logic in this PR

Nuru:
@woz5999 Note that #32 is open and likely to get merged very soon, so maybe wait until that is released to rebase your PR. Please close your PR if you are satisfied with using v0.11.1 when it comes out.

woz5999:
@Nuru modified the PR such that it only adds the ability for the users to attach additional security groups to the generated launch template. it also fixes a bug wherein `remote_access` and `launch_template` are mutually exclusive on the `eks_node_group` resource.

Nuru (approved):

PR #36, opened 2020-10-08
feat: allow ebs launch template encryption

### LINKED ISSUE #35: EBS Node Encryption
## Describe the Feature

Support [EBS Encryption](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#block-devices) via the `ebs` block when a `launch_template` is used

## Expected Behavior

When the user sets a flag or field, they should be able get EBS volumes attached the nodes via the launch template to be encrypted by default.

## Use Case

When using the managed EKS Node Group, I need to be able adhere to certain compliance standards, which includes asserting that all data is encrypted at rest.  I am unable to achieve said compliance constraint if I cannot get EBS encrypted volumes in my node group.

## Describe Ideal Solution

The launch template created [here](https://github.com/cloudposse/terraform-aws-eks-node-group/blob/master/launch-template.tf#L48) could use the [ebs.encrypted](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#encrypted) flag, and surface that on this module as a variable like: `launch_template_ebs_encryption: boolean`

## Alternatives Considered

None

dotCipher:
## what
* Surface variable for boolean flag of `launch_template_disk_encryption`
* Use `launch_template_disk_encryption` to flip flag of generated `launch_template.ebs.encryption`

## why
* Allow EBS encryption 

## references
* Closes #35

### DIFF
--- a/launch-template.tf
+++ b/launch-template.tf
@@ -62,7 +62,7 @@ resource "aws_launch_template" "default" {
 
     ebs {
       volume_size = var.disk_size
-      encrypted = var.launch_template_disk_encryption
+      encrypted   = var.launch_template_disk_encryption
     }
   }
 
--- a/launch-template.tf
+++ b/launch-template.tf
@@ -62,7 +62,7 @@ resource "aws_launch_template" "default" {
 
     ebs {
       volume_size = var.disk_size
-      encrypted   = var.launch_template_disk_encryption
+      encrypted   = var.launch_template_disk_encryption_enabled
     }
   }
 
--- a/launch-template.tf
+++ b/launch-template.tf
@@ -62,6 +62,7 @@ resource "aws_launch_template" "default" {
 
     ebs {
       volume_size = var.disk_size
+      encrypted = var.launch_template_disk_encryption
     }
   }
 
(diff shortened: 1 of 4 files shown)


variables.tf
```diff
@@ -185,6 +185,12 @@ variable "module_depends_on" {
   description = "Can be any value desired. Module will wait for this value to be computed before creating node group."
 }
 
+variable "launch_template_disk_encryption" {
+  type        = bool
+  description = "Enable disk encryption if not providing `launch_template_name` of an existing template."
```
aknysh, line 190:
```suggestion
  description = "Enable disk encryption for the created launch template (if we aren't provided with an existing launch template)"
```

aknysh (commented):
see comments

dotCipher:
Cool, thanks for the review!

launch-template.tf
```diff
@@ -62,6 +62,7 @@ resource "aws_launch_template" "default" {
 
     ebs {
       volume_size = var.disk_size
+      encrypted   = var.launch_template_disk_encryption
```
aknysh, line 65:
```suggestion
      encrypted   = var.launch_template_disk_encryption_enabled
```

aknysh (changes requested):
please see comments

launch-template.tf
dotCipher: Ah right my bad :)

aknysh (approved):
thanks @dotCipher

PR #44, opened 2020-11-30
add disk options for ec2 template

dmitriy-lukyanchikov:
## what
- add options to define volume_type and km key for volume

## why
- its obvious that sometimes we need to define it


dmitriy-lukyanchikov:
@aknysh please review this one

Nuru (approved):

PR #45, opened 2020-12-04
Add `capacity_type` to support node group spot pricing

ChrisMcKee:
## what
* Adds in support for capacity_type settings on node group 
* Removes clrf from document https://www.atechtown.com/remove-ctrl-m-character/
* Adds gitattributes so linux line endings are maintained

## why
* Who doesn't like spot pricing 
* I don't like paying full price for machines

## references
* https://aws.amazon.com/blogs/containers/amazon-eks-now-supports-provisioning-and-managing-ec2-spot-instances-in-managed-node-groups/
* https://aws.amazon.com/about-aws/whats-new/2020/12/amazon-eks-support-ec2-spot-instances-managed-node-groups/
* https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group#capacity_type

PR #54, opened 2021-01-20
Fix allow several instance types

patrickjahns:
## what
- Allow to specify more than a single instance_types for eks node groups
- Possible to pass more than a single item in the list since the the introduction of `capacity_type` [see upstream pr](https://github.com/hashicorp/terraform-provider-aws/pull/16510)

## why
- For utilising SPOT instances it is a good practice to specify more than a single instance type

## references
- Related to #8
- See also `instanceTypes` in [current AWS eks api](https://docs.aws.amazon.com/eks/latest/APIReference/API_CreateNodegroup.html#API_CreateNodegroup_RequestSyntax)


patrickjahns:
I'll take a look later at fixing the tests - seems to be that they expect the validation message to be in a certain format.
Anything else I should consider ?

SweetOps:
> Anything else I should consider ?

fix of `validation message` should be enough

patrickjahns:
/test all

🤞 - hope this works

Nuru (changes requested):
@osterman @aknysh We need to discuss how to support this feature moving forward.

One problem is that supporting this feature [requires AWS Terraform provider >= 3.20](https://github.com/hashicorp/terraform-provider-aws/issues/16547#issuecomment-738450595) and I do not know if we want to push that requirement on our customers. At a minimum, though, we need to make that change in [versions.tf](https://github.com/cloudposse/terraform-aws-eks-node-group/blob/b4c6e027c4d281e5dd557ec2b10b9c5542c6b925/versions.tf#L7) as part of this PR.

A bigger problem is that AWS is currently split on how they support this, which brings us a split in how we can support it. I would prefer to defer support until AWS resolves their split, but I can see where our customers would get upset about that.

According to the [AWS documentation](https://docs.aws.amazon.com/eks/latest/APIReference/API_CreateNodegroup.html#API_CreateNodegroup_RequestSyntax):
> If you specify launchTemplate, then you can specify zero or one instance type in your launch template or you can specify 0-20 instance types for instanceTypes.

The problem for us with this module is that there are a set of features only available by specifying the launch template: See https://github.com/cloudposse/terraform-aws-eks-node-group/blob/b4c6e027c4d281e5dd557ec2b10b9c5542c6b925/main.tf#L9

This module automatically creates a launch template if the requested configuration requires it, which means that it is not obvious whether or not you can specify more than one instance type, and makes error messages confusing. (I think the test for whether or not we can have more than once instance type is more complex than allowed in a custom validation. If I am wrong about that, then I guess a custom validation with appropriate explanation would be they way to move forward.)

patrickjahns:
@Nuru 
> One problem is that supporting this feature requires AWS Terraform provider >= 3.20 and I do not know if we want to push that requirement on our customers. At a minimum, though, we need to make that change in versions.tf as part of this PR.

The support for `capacity_type` was already added in https://github.com/cloudposse/terraform-aws-eks-node-group/pull/45 - so my assumption was, that this requirement for the provider version was already "approved"

Regarding:
> If you specify launchTemplate, then you can specify zero or one instance type in your launch template or you can specify 0-20 instance types for instanceTypes.

The next sentences from the AWS documentation clarified it the launch template situation for me
> If however, you specify an instance type in your launch template and specify any instanceTypes, the node group deployment will fail

So I understood this, as that you are limited to 0-1 instance-types when you specify it via the launch template - but if you specify the instance types via the `instance_types` in the  `aws_eks_node_group` resource it should not be a problem.

For users of the module, it is probably good to raise awareness, that when they use a custom launch template they shall rather omit the `instance_type` in the launch template to be able to specify more than a single type for a node group

patrickjahns:
Anything I can do here to move forward with this PR?

variables.tf
```diff
@@ -106,14 +106,14 @@ variable "instance_types" {
...
   description = <<-EOT
-    Single instance type to use for this node group, passed as a list. Defaults to ["t3.medium"].
-    It is a list because Launch Templates take a list, and it is a single type because EKS only supports a single type per node group.
+    Instance types to use for this node group (up to 20). Defaults to ["t3.medium"].
+    Ignored when `launch_template_id` is supplied.
     EOT
   validation {
     condition = (
-      length(var.instance_types) == 1
+      length(var.instance_types) > 20
```
Nuru, line 114: > The condition argument is an expression that must use the value of the variable to return true if the value is valid, or false if it is invalid.

```suggestion
      length(var.instance_types) >= 1 && length(var.instance_types) <= 20
```

Nuru:
One remaining issue is that this module creates launch templates implicitly for the user to enable certain features. We now need to update the module so that the launch template does NOT [specify instance types](https://github.com/cloudposse/terraform-aws-eks-node-group/blob/caf738c31aea9a503e69e82aa4aa2927bf38ee34/launch-template.tf#L74) and ensure that the `aws_eks_node_group` ALWAYS [specifies instance types](https://github.com/cloudposse/terraform-aws-eks-node-group/blob/caf738c31aea9a503e69e82aa4aa2927bf38ee34/main.tf#L96).

Nuru (approved):

PR #68, opened 2021-05-25
feat: use security-group module instead of resource

SweetOps:
## what
* use security-group module instead of resource
* update tests

## why
* more flexible  than current implementation 
* bring configuration of security group/rules to one standard

## references
* CPCO-409


outputs.tf
```diff
@@ -29,6 +29,16 @@ output "eks_node_group_status" {
...
+
+output "eks_node_group_remote_access_security_group_arn" {
+  description = "ARN of the EKS cluster Security Group for remote access to EKS Node Group"
+  value       = module.security_group.arn
+}
+
+output "eks_node_group_remote_access_security_group_name" {
+  description = "Name of the EKS cluster Security Group for remote access to EKS Node Group"
+  value       = module.security_group.name
 }
```
nitrocode, line 44: Also (or in place of additional outputs) output the entire `module.security_group`

```suggestion
}

output "eks_node_group_remote_access_security_group" {
  description = "EKS cluster Security Group for remote access to EKS Node Group"
  value       = module.security_group
}
```

nitrocode (approved):

PR #66, opened 2021-04-16
Ability to add additional security groups to a node group

nustiueudinastea:
## what
* Adds the ability to assign additional security groups to a node group using a module variable.

## why
* It's useful to be able to add extra security groups with custom user rules.
* The workers module also supports this feature (https://registry.terraform.io/modules/cloudposse/eks-workers/aws/latest)

PR #63, opened 2021-03-15
feat(aws_launch_template): parametrize metadata_options

abhinavkhanna-sf:
## what
* parametrized launch template metadata options 
* moved default values to variables.tf

## why
* Restricting access to the IMDS and Amazon EC2 instance profile credentials
* [EKS Security Best Practices](https://docs.aws.amazon.com/eks/latest/userguide/best-practices-security.html)

## references
* closes #62

PR #71, opened 2021-06-29
add variable for http_put_response_hop_limit

tarfik:
what
changes from https://github.com/cloudposse/terraform-aws-eks-node-group/pull/66
and
get http_put_response_hop_limit from variables

why
If we need to disable metadata API for pods without serviceaccount and network!=host we need to set http_put_response_hop_limit == 1.

references
https://docs.aws.amazon.com/eks/latest/userguide/best-practices-security.html#restrict-ec2-credential-access


Nuru:
@tarfik Thank you for this contribution. This is a difficult module to work with and you did well. 

I do not understand the dates and timing of this PR exactly, but the feature you requested regarding http_put_response_hop_limit was added in #63 and released in v0.20.0 of this module.

Changes from #66 will be addressed in that PR.

Closing this PR as duplicate/not needed.

Issue #72, opened 2021-07-20
Make device_name of block_device_mappings parameterised


lyubomirdimov:
The launch template is currently creating a static block device mapping 
``` 
block_device_mappings {
    device_name = "/dev/xvda"

    ebs {
      volume_size = var.disk_size
      volume_type = var.disk_type
      kms_key_id  = var.launch_template_disk_encryption_enabled && length(var.launch_template_disk_encryption_kms_key_id) > 0 ? var.launch_template_disk_encryption_kms_key_id : null
      encrypted   = var.launch_template_disk_encryption_enabled
    }
  }
```

I am working with Bottlerocket (https://github.com/bottlerocket-os/bottlerocket) and it requires different device_name for its block_device_mappings. 

Would it be possible to make the launch template a bit more configurable?

Should I be creating a custom launch template instead?

Update:
Open PR - https://github.com/cloudposse/terraform-aws-eks-node-group/pull/73

Update 2:
Closed the PR - It is insufficient for the use case. We need a functionality in which we can create multiple device mappings for the launch template

z0rc:
@lyubomirdimov What is end goal here?

As I'm reading https://github.com/bottlerocket-os/bottlerocket#default-volumes, it actually requires two volumes. So maybe just changing single volume name isn't enough and we need ability to configure whole `block_device_mappings {}`?

lyubomirdimov:
@z0rc you are right, we will need a way to create multiple mappings (if tf allows a for loop within the `block_device_mappings {}` ). Thus my PR is insufficient for the Bottlerocket case and I will close it for now.

z0rc:
@lyubomirdimov can you please open new ticket about multiple `block_device_mappings {}` blocks, so this topic won't be lost?

lyubomirdimov:
@z0rc 
Yes I will open an issue later today. I was also working on a PR for, with `dynamic` object which creates multiple mappings.

lyubomirdimov:
Issue - https://github.com/cloudposse/terraform-aws-eks-node-group/issues/74
PR - https://github.com/cloudposse/terraform-aws-eks-node-group/pull/75

PR #73, opened 2021-07-20
feat: parametarize device_name of the block_device_mapping created by…

lyubomirdimov:
Launch template block device mappings - make `device_name` parametarized. 

## what
* introducing new variable `launch_template_disk_device_name`, which defaults to `/dev/xvda` (default for Amazon Linux 2 AMI)
* Use `launch_template_disk_device_name` as input for `device_name` in the launch template

## why
* When using custom AMI (f.ex. Bottlerocket) sometimes we need to map to storage device different than `/dev/xvda`

## references
* I have opened an Issue https://github.com/cloudposse/terraform-aws-eks-node-group/issues/72


lyubomirdimov:
Closing the PR, it is insufficient for the use case. We need a functionality in which we can create multiple device mappings for the launch template

PR #77, opened 2021-07-27
Always add var.security_groups to launch template if provided

cvittoriasona:
## what
* If `var.security_groups` is present, add any passed in security groups, along with the default cluster security group, to the launch template.

## why
* `var.security_groups` is only added to the launch template if `var.remote_access_enabled` is true. Additional security groups should not be dependent on SSH access being enabled to be used.
* Specifically, ran into an issue when using a x-account shared VPC where the default security group for the VPC was not available to accounts the VPC was shared with. After encountering this error, when attempting to specify a security group for the launch template using `var.security_groups`, realized this var isn't active unless `var.remote_access_enabled` is also set. See below for output:
```
Error: error creating EKS Node Group (my-eks-node-group): InvalidRequestException: You do not have access to a default security group in VPC vpc-123456. Specify a security group, 310. Specify a security group, and try again.
│ {
│   RespMetadata: {
│     StatusCode: 400,
│     RequestID: "some-uuid"
│   },
│   Message_: "You do not have access to a default security group in VPC vpc-123456. Specify a security group, and try again."
│ }
```

This seems to be mostly a workaround for launch templates as EKS managed nodegroups should be auto-assigned to the default cluster security group, even if the launch template has no security groups attached to it.

Issue was present in v0.19.0 only when using `var.kubernetes_taints`, but in >=v0.20.0 this issue applied to all nodegroups created with this module.

## references
* Tested with AWS provider v3.44.0 & v3.50.0


jamengual:
@cvittoriasona do not use shared vpc, they suck.

jamengual (approved):

PR #90, opened 2021-09-30
add custom_ami_regex variable

mcalhoun:
## what

* Add `custom_ami_regex` variable

## why

* Allow the consumer to specify a custom AMI regex for the worker nodes


mcalhoun:
After further investigation, this change isn't necessary.

PR #103, opened 2022-01-12
Output launch template name

### LINKED ISSUE #102: Output aws_launch_template.default.name
## Describe the Feature

Add `aws_launch_template_default.name` to outputs.tf.

## Expected Behavior

Results in the ability to obtain the launch template name created by this module.

## Use Case

If an autoscaling tool like `karpenter` is used for eks workers, this module should return the launch template name created for the workers in the node group so that it can be passed to `karpenter` to ensure consistency across nodes.

https://karpenter.sh/docs/aws/provisioning/#launchtemplate

bwmetcalf:
## what
Adds launch template name to outputs

## why
* This is useful for cases where a tool like `karpenter` is used to scale nodes and the same launch template should be used.

## references
* closes #102

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -40,5 +40,5 @@ output "eks_node_group_cbd_pet_name" {
 
 output "eks_worker_launch_template_name" {
   description = "The launch template name provided or created for the worker nodes"
-  value       = local.fetch_launch_template ? data.aws_launch_template.this.name : aws_launch_template.default.name
+  value       = local.enabled ? (local.fetch_launch_template ? data.aws_launch_template.this[0].name : aws_launch_template.default[0].name) : ""
 }
--- a/outputs.tf
+++ b/outputs.tf
@@ -39,6 +39,6 @@ output "eks_node_group_cbd_pet_name" {
 }
 
 output "eks_worker_launch_template_name"
-  description = "The launch template name created for the worker nodes, if this module generated one"
-  value       = join("", aws_launch_template.*.name)
+  description = "The launch template name provided or created for the worker nodes"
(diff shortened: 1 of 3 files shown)


outputs.tf
```diff
@@ -37,3 +37,8 @@ output "eks_node_group_cbd_pet_name" {
   description = "The pet name of this node group, if this module generated one"
   value       = join("", random_pet.cbd.*.id)
 }
+
+output "eks_worker_launch_template_name" {
+  description = "The launch template name provided or created for the worker nodes"
+  value       = local.fetch_launch_template ? data.aws_launch_template.this.name : aws_launch_template.default.name
```
aknysh, line 43:
```suggestion
  value       = local.enabled ? (local.fetch_launch_template ? data.aws_launch_template.this[0].name : aws_launch_template.default[0].name) : ""
```

Since all those resources have `count`, they are already lists (not single items), to to get an output from them we need to use the list notation `[0]`

aknysh (changes requested):
thanks @bwmetcalf  please see comments

outputs.tf
```diff
@@ -37,3 +37,8 @@ output "eks_node_group_cbd_pet_name" {
   description = "The pet name of this node group, if this module generated one"
   value       = join("", random_pet.cbd.*.id)
 }
+
+output "eks_worker_launch_template_name" {
+  description = "The launch template name provided or created for the worker nodes"
+  value       = local.enabled ? (local.fetch_launch_template ? data.aws_launch_template.this[0].name : aws_launch_template.default[0].name) : ""
```
nitrocode, line 43:
```suggestion
  value       = local.enabled ? (local.fetch_launch_template ? data.aws_launch_template.this[0].name : aws_launch_template.default[0].name) : null
```

PR #84, opened 2021-08-29
Overhaul

Nuru:
## what
Major overhaul. See [MIGRATION.md](https://github.com/cloudposse/terraform-aws-eks-node-group/blob/overhaul/MIGRATION.md) for details.

- Update to new Cloud Posse [security group](https://github.com/cloudposse/terraform-aws-security-group) module and variable naming standards
- Incorporate features from all open PRs. Supersedes and
  - closes #51 
  - closes #65 
  - closes #66 
  - closes #75 
  - closes #78 
- Bug fixes:
  - fixes #37 
  - fixes #64 
  - fixes #67
  - fixes #69
  - fixes #74 
- Update Cloud Posse standard workflows

## why
- Take advantage of recently added features in both EKS and Terraform
- Improve usability
- Standardize

PR #105, opened 2022-01-19
Update pet name when changing launch template id

woz5999:
## what
* Update pet name when changing launch template id

## why
* If you change the template id, tf/aws will try to create a new node group; however, because we're using create before destroy, it errors out due to "NodeGroup already exists with name food and cluster name bar"
* generating a new pet name will restore the correct behavior
* note that this was the behavior prior to https://github.com/cloudposse/terraform-aws-eks-node-group/pull/84


context.tf
```diff
@@ -20,6 +20,7 @@
 # will be null, and `module.this.delimiter` will be `-` (hyphen).
 #
 
+
```
aknysh, line 23: why this file has been changed?
It should be taken from https://github.com/cloudposse/terraform-null-label/blob/master/exports/context.tf

examples/complete/context.tf
```diff
@@ -20,6 +20,7 @@
 # will be null, and `module.this.delimiter` will be `-` (hyphen).
 #
 
+
```
aknysh, line 23: why this file has been changed?
It should be taken from https://github.com/cloudposse/terraform-null-label/blob/master/exports/context.tf

woz5999: ah sorry. i see a pushed of pollution from when i updated my fork. i'll fix everything up nice. i'll set and unset draft status accordingly.

PR #106, opened 2022-01-19
Update pet name when changing launch template id

woz5999:
## what
* Update pet name when changing launch template id

## why
* If you change the template id, tf/aws will try to create a new node group; however, because we're using create before destroy, it errors out due to "NodeGroup already exists with name food and cluster name bar"
* generating a new pet name will restore the correct behavior
* note that this was the behavior prior to Overhaul #84


woz5999:
@aknysh i ended up just creating a new fork since mine was so old. here's the fixed version of !105

woz5999:
anyone available to review this?

woz5999:
@aknysh @Nuru @Gowiem is anyone able to review this?

Gowiem:
Since @Nuru did #84, I'll leave it to him.

PR #116, opened 2022-05-15
Update main.tf

### LINKED ISSUE #123: after_cluster_joining_userdata only allows one element
## Describe the Bug

```
  type        = list(string)
  default     = []
  validation {
    condition = (
    )
  }
}
```

## Expected Behavior

surajrubik:
Idea is to have the name in ec2 instance i.e is created by the autoscalling group.

## what
* Describe high-level what changed as a result of these commits (i.e. in plain-english, what do these changes mean?)
* Use bullet points to be concise and to the point.

## why
* Provide the justifications for the changes (e.g. business case). 
* Describe why these changes were made (e.g. why do these commits fix the problem?)
* Use bullet points to be concise and to the point.

## references
* Link to any supporting github issues or helpful documentation to add some context (e.g. stackoverflow). 
* Use `closes #123`, if this PR closes a GitHub issue `#123`

### DIFF
--- a/.github/workflows/auto-readme.yml
+++ b/.github/workflows/auto-readme.yml
@@ -1,5 +1,7 @@
 name: "auto-readme"
 on:
+  workflow_dispatch:
+
   schedule:
   # Example of job definition:
   # .---------------- minute (0 - 59)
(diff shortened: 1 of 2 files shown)


Gowiem:
@surajrubik I think I get what you're trying to do with this PR, but that is not how this should be accomplished. 

We supply `Name` via `module.label.tags` which is the base of `node_tags`. This would override that and create an undesirable result. 

Instead, you can pass the value of `cluster_name` into the `tags` input of the module as a new `ClusterName` tag and it will result in being applied to all resources. Can you please try that and let me know how it works for you? Until then, I'm going to close this out as I believe this would break other's existing tag structures erroneously. Thanks!

surajrubik:
@Gowiem, did get you. I am not finding any variable to pass the same that you said, would help me to find it, please.

Gowiem:
@surajrubik try this input to the module: https://github.com/cloudposse/terraform-aws-eks-node-group#input_tags

PR #112, opened 2022-03-02
feat: add tags_all to the output

abhinavkhanna-sf:
## what
adds tags_all to the output

## why
* missing output from the module
* tags_all can be used as input to tag ASG

## references
* closes #111

PR #118, opened 2022-05-21
Cleanup for V2 release

Nuru:
## what

- Add updated documentation for migration to V2.0.0
- Change default for `ebs_optimized` to `true`
- Add tags to the output, supersedes and closes #112
- Fix the tag value for Auto Scaler tags that indicate taints

## why

- Convert to production Semantic Versioning 
- Best practice is to set `ebs_optimized` to `true` and it is the default for most instances, so it was a mistake to default it to `false` when the input was introduced
- Requested feature
- Taint effect was not being properly indicated


docs/migration-v1-v2.md
```diff
@@ -0,0 +1,218 @@
...
+
+In most cases, the changes you need to make are pretty easy.
+
+#### Review behavior changes and new features
+
+- Do you want node group instance EBS volumes deleted on termination? You can disable that now.
+- Do you want Instance Metadata Service v1 available? This module now disables it by default, and EKS and Kubernetes all handle that fine, but you might have scripts that `curl` the instance
+  metadata endpoint that need it.
+- Did you have the "create before destroy" behavior disabled? The migration to v2.0.0 of this module is going to cause your node group to be destroyed and recreated anyway, so take the
+  opportunity to enable it. It will save you and outage some day.
```
aknysh, line 128:
```suggestion
  opportunity to enable it. It will save you an outage some day.
```

test/src/go.mod
```diff
@@ -1,47 +1,46 @@
-module github.com/cloudposse/terraform-aws-eks-node-group
+module github.com/cloudposse/terraform-aws-eks-cluster
```
aknysh, line 1: this should be `github.com/cloudposse/terraform-aws-eks-node-group`

main.tf
```diff
@@ -13,9 +13,16 @@ locals {
 
   get_cluster_data = local.enabled ? (local.need_cluster_kubernetes_version || local.need_bootstrap || local.need_ssh_access_sg || length(var.associated_security_group_ids) > 0) : false
 
+  taint_effect_map = {
```
aknysh, line 16: why this is necessary?
why the use can't use `NoSchedule` instead of `NO_SCHEDULE`?

aknysh (commented):
a few nitpicks

main.tf
Nuru: Because we need `NO_SCHEDULE` in one place and `NoSchedule` in another based on the same input which is already documented to require `NO_SCHEDULE`.

aknysh (approved):

PR #119, opened 2022-05-26
Update pet name for new launch template version

bwmetcalf:
## what
* Update pet name for new launch template version to force a new node group.

## why
* A new pet name and therefore node group needs to be created when the launch template version changes.

## references
* This is similar to #106

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -102,7 +102,8 @@ resource "random_pet" "cbd" {
     ami_type       = local.ng.ami_type
     capacity_type  = local.ng.capacity_type
 
-    launch_template_id = local.launch_template_id
+    launch_template_id      = local.launch_template_id
+    launch_template_version = local.launch_template_version
   }
 }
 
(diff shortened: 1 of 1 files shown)


bwmetcalf:
@nitrocode Could you help move this along?  Thanks.

nitrocode:
What if you do not want the node group to recreate if you update the launch template version ? We update the ami on the launch template all the time and we'd prefer not recreating the node group. I would much rather update our existing node groups after updating the launch template using this module.

https://docs.aws.amazon.com/eks/latest/userguide/update-managed-node-group.html

> ## Updating a managed node group
>
> * A new AMI release version is available for your managed node group. For more information about AMI versions, see these sections:

> eksctl upgrade nodegroup --name=node-group-name --cluster=cluster-name

Nuru:
Closing, because we do not want to force a new node group if we do not have to. You can use `terraform taint` if you want to force a new node group, but you probably want to use [instance refresh](https://aws.amazon.com/blogs/compute/introducing-instance-refresh-for-ec2-auto-scaling/) instead to simply update the node group rather than create a new one.

Issue #95, opened 2021-11-05
Support attribute-based instance type selection


Nuru:
Support attribute-based instance type selection when the AWS Terraform provider supports it.
See [AWS announcement](https://aws.amazon.com/about-aws/whats-new/2021/10/amazon-ec2-instance-auto-scaling-ec2-spot-fleet/)

nitrocode:
Ticket to track https://github.com/hashicorp/terraform-provider-aws/issues/21566

jtdoepke:
> Ticket to track https://github.com/hashicorp/terraform-provider-aws/issues/21566

This has been closed/released in in [AWS Provider v4.16.0](https://github.com/hashicorp/terraform-provider-aws/blob/v4.16.0/CHANGELOG.md)

nitrocode:
@jtdoepke cool, thanks for the heads up!

Looks like we can use the `instance_requirements` block on the launch template.

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#instance-requirements

PR #120, opened 2022-06-07
Support attribute-based instance type selection

nitrocode:
## what
* Expose instance_requirements
* This is a WIP

## why
* Support attribute-based instance type selection

## references
* Potentially closes (or will eventually close) https://github.com/cloudposse/terraform-aws-eks-node-group/issues/95
* block description https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#instance_requirements
* full block definition https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#instance-requirements
* t3 info https://aws.amazon.com/ec2/instance-types/t3/
* Perhaps we want to match https://github.com/terraform-aws-modules/terraform-aws-autoscaling/blob/e14d875cc2ee8a5c2882ed41df0d8135cbcf530b/main.tf#L159 ?

## test

```hcl
module "eks_node_group" {
  # Use source to point to this branch
  # NOTE: this branch is experimental and will be removed upon merge
  source = "git::https://github.com/cloudposse/terraform-aws-eks-node-group?ref=instance_requirements"

  # should select t3.medium (untested)
  instance_requirements = {
    memory_mib = {
      min = 4
      max = 4
    }
    memory_gib_per_vcpu = {
      min = 0.5
      max = 0.5
    }
    vcpu_count = {
      min = 2
      max = 2
    }
    cpu_manufacturers    = ["amazon-web-services"]
    instance_generations = ["current"]
  }
  # ...
}
```


jtdoepke:
I think the provider version needs to be bumped to`>= 4.14.0` because [that's when `instance_requirements` was added to launch templates](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md#4140-may-13-2022).

Also, maybe `instance_requirements` should be part of the [create_before_destroy pet name](https://github.com/cloudposse/terraform-aws-eks-node-group/blob/179c7d9b13c68cff4b5b776f58f09fca80f4fe99/main.tf#L101)?

nitrocode:
> Also, maybe instance_requirements should be part of the [create_before_destroy pet name](https://github.com/cloudposse/terraform-aws-eks-node-group/blob/179c7d9b13c68cff4b5b776f58f09fca80f4fe99/main.tf#L101)?

I don't think the `instance_requirements` should be part of that because then this will auto create the node group which would destroy the existing one. The `instance_requirements` will only affect the launch template and then we can approach rotating the node group safely outside of terraform.

nitrocode:
@jtdoepke see the above `test` in the PR description. Please test this out if you have time. :)

awoimbee:
This PR won't work, EKS managed node groups are not compatible with `instance_requirements`. What will happen here is that AWS/EKS will spawn t3.medium nodes only

PR #122, opened 2022-07-12
Change validation of put response hop limit to allow `1` as value to limit access to worker node's metadata endpoint

jakubbujny:
## what
* Change validation of `metadata_http_put_response_hop_limit` variable to allow to set `1` as value.

## why
* as standing in EKS best practises to limit access to worker node's metadata endpoint it's required to enable `metadata_http_tokens_required` and set `metadata_http_put_response_hop_limit` to 1 - see https://aws.github.io/aws-eks-best-practices/security/docs/iam/#restrict-access-to-the-instance-profile-assigned-to-the-worker-node

```
aws ec2 modify-instance-metadata-options --instance-id <value> --http-tokens required --http-put-response-hop-limit 1
```


variables.tf
```diff
@@ -367,13 +367,12 @@ variable "metadata_http_put_response_hop_limit" {
   default     = 2
   description = <<-EOT
     The desired HTTP PUT response hop limit (between 1 and 64) for Instance Metadata Service requests.
-    The default is `2` to support containerized workloads.
```
Gowiem, line 370: Let's revert this line delete back unless you've got a reason for it. I see no reason to remove this additional bit of info.

Gowiem (changes requested):

variables.tf
jakubbujny: @Gowiem done

christoffer-eide:
Any hope of getting this PR merged?

PR #136, opened 2022-12-21
Prevent unexpected privileges escalation

gillg:
## what
The current variable `input_metadata_http_put_response_hop_limit` condition, prevent to protect users of this module, to be protected against privileges escalation.
The first intent of IMDSv2 is to prevent containers beeing able to assume an EC2 instance profile. It's not a bad idea at all to prevent that. The good practice then is to use the module `cloudposse/eks-iam-role/aws` to create a kubernetes service account mapped with IAM permissions throug an OIDC IdP.

## references
* https://aws.github.io/aws-eks-best-practices/security/docs/iam/#restrict-access-to-the-instance-profile-assigned-to-the-worker-node


gillg:
Does anyone can take a look ? randomly @Gowiem

max-lobur (approved):

max-lobur:
Makes sense, thanks for this. I think we will change default too, but will check regression first

gillg:
> Makes sense, thanks for this. I think we will change default too, but will check regression first

Fine to me, but be careful with the default... ^^ A lot of people will have unexpected permissions not working anymore. It's a breaking change.

max-lobur:
Both failures unrelated, I will fix in separate PR

xeivieni:
Any update on this PR ? I would love to see this released 🤗

nitrocode:
@max-lobur this is odd. It seems like it's trying to revert it back to 2022 instead of using 2023.

It should be using the latest build-harness which grabs rhe current year.

https://github.com/cloudposse/build-harness/blob/c7b01d772017cc78b3902c0a7fcd54f6b6131dca/templates/README.md.gotmpl#L288

nitrocode:
Oh I see the issue. We just need to run the auto format again on this pr since this pr was submitted in 2022. Now its 2023 and the year should be 2023 in the readme.

Please run `make pr/auto-format` locally

nitrocode (dismissed):

examples/complete/main.tf
```diff
@@ -117,7 +117,7 @@ module "https_sg" {
 
 module "eks_cluster" {
   source  = "cloudposse/eks-cluster/aws"
-  version = "2.2.0"
+  version = "2.6.0"
```
nitrocode, line 120:
```suggestion
  version = "2.4.0"
```

nitrocode (approved):

nitrocode:
Thank you @gillg for the contribution!

PR #137, opened 2022-12-21
Add missing managed AMI codes

gillg:
## what
* Add some missing images codes from https://docs.aws.amazon.com/eks/latest/APIReference/API_Nodegroup.html

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -111,14 +111,14 @@ variable "ami_type" {
   type        = string
   description = <<-EOT
     Type of Amazon Machine Image (AMI) associated with the EKS Node Group.
-    Defaults to `AL2_x86_64`. Valid values: `AL2_x86_64`, `AL2_x86_64_GPU`, `AL2_ARM_64`, `BOTTLEROCKET_x86_64`, and `BOTTLEROCKET_ARM_64`.
+    Defaults to `AL2_x86_64`. Valid values: https://docs.aws.amazon.com/eks/latest/APIReference/API_Nodegroup.html.
     EOT
   default     = "AL2_x86_64"
   validation {
     condition = (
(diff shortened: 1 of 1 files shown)


variables.tf
```diff
@@ -111,14 +111,14 @@ variable "ami_type" {
   type        = string
   description = <<-EOT
     Type of Amazon Machine Image (AMI) associated with the EKS Node Group.
-    Defaults to `AL2_x86_64`. Valid values: `AL2_x86_64`, `AL2_x86_64_GPU`, `AL2_ARM_64`, `BOTTLEROCKET_x86_64`, and `BOTTLEROCKET_ARM_64`.
+    Defaults to `AL2_x86_64`. Valid values: https://docs.aws.amazon.com/eks/latest/APIReference/API_Nodegroup.html.
```
nitrocode, line 114:
```suggestion
    Defaults to `AL2_x86_64`. Valid values are available in the [AWS API Node Group docs](https://docs.aws.amazon.com/eks/latest/APIReference/API_Nodegroup.html).
```

nitrocode:
Please fill out the full PR template

variables.tf
```diff
@@ -111,14 +111,14 @@ variable "ami_type" {
   type        = string
   description = <<-EOT
     Type of Amazon Machine Image (AMI) associated with the EKS Node Group.
```
nitrocode, line 113: Have you tested this change?

I think you may need more changes to this file

https://github.com/cloudposse/terraform-aws-eks-node-group/blob/ea8401ef0db1cce91fd96ce1776407f552dbc090/ami.tf#L4

gillg: Indeed I missed the file.
I not tested properly, I was in rush working locally with quick tricks to make my need work.

I will take a look better, thanks for catching that!

PR #139, opened 2023-01-03
Windows node support

ChrisMcKee:
## what
* Adds support for 2019/2022 (windows lts) ami-types as per the aws sdk docs
* Docs to support usage and link to ref docs https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html

## why
* Windows EKS Managed node support https://github.com/aws/containers-roadmap/issues/584

## references
* https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html
* https://docs.aws.amazon.com/eks/latest/userguide/eks-ami-versions-windows.html

## Tested 
image
image
image
image

```hcl
module "eks_windows_node_group" {
  # source  = "cloudposse/eks-node-group/aws"
  # version = "2.6.1"
  source = "github.com/ChrisMcKee/terraform-aws-eks-node-group"

  instance_types     = ["t3.large", "t3a.large", "c5.large", "c6i.large", "m6i.large", "r6i.large"]
  subnet_ids         = [data.terraform_remote_state.network.outputs.private_subnets[1]]
  min_size           = 1
  max_size           = 1
  desired_size       = 1
  cluster_name       = module.eks_cluster.eks_cluster_id
  kubernetes_version = var.kubernetes_version == null || var.kubernetes_version == "" ? [] : [var.kubernetes_version]
  kubernetes_labels  = var.labels

  ami_type = "WINDOWS_CORE_2019_x86_64"

  update_config = [{ max_unavailable = 1 }]

  capacity_type = "SPOT"

  kubernetes_taints = [{
    key    = "WINDOWS"
    value  = "true"
    effect = "NO_SCHEDULE"
  }]

  node_role_arn                = [aws_iam_role.worker_role_nt.arn]
  node_role_cni_policy_enabled = false #We use the Service Account as per best practice

  associated_security_group_ids = [data.terraform_remote_state.network.outputs.ops_ssh, aws_security_group.workers.id]

  # Enable the Kubernetes cluster auto-scaler to find the auto-scaling group
  cluster_autoscaler_enabled = true

  context = module.windowslabel.context

  # Ensure the cluster is fully created before trying to add the node group
  module_depends_on = [module.eks_cluster.kubernetes_config_map_id]

  # Ensure ordering of resource creation to eliminate the race conditions when applying the Kubernetes Auth ConfigMap.
  # Do not create Node Group before the EKS cluster is created and the `aws-auth` Kubernetes ConfigMap is applied.
  depends_on = [module.eks_cluster, module.eks_cluster.kubernetes_config_map_id]

  create_before_destroy = true

  node_role_policy_arns = ["arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore"]

  block_device_mappings = [
    {
      "delete_on_termination" : true,
      "device_name" : "/dev/xvda",
      "encrypted" : true,
      "volume_size" : 80,
      "volume_type" : "gp3"
    }
  ]

  node_group_terraform_timeouts = [{
    create = "40m"
    update = null
    delete = "20m"
  }]

  #Valid types are "instance", "volume", "elastic-gpu", "spot-instances-request", "network-interface".
  resources_to_tag = ["instance", "volume", "spot-instances-request", "network-interface"]
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,31 @@ Available targets:
 ```
 <!-- markdownlint-restore -->
 <!-- markdownlint-disable -->
+## Windows Managed Node groups
+  Windows managed node-groups have a few pre-requisites.
+
+  * Your cluster must contain at least one linux based worker node
+  * Your EKS Cluster must have the `AmazonEKSVPCResourceController` and `AmazonEKSClusterPolicy` policies attached
+  * Your cluster must have a config-map called amazon-vpc-cni with the following content
+  ```yaml
+  apiVersion: v1
+  kind: ConfigMap
+  metadata:
+  name: amazon-vpc-cni
+  namespace: kube-system
+  data:
+  enable-windows-ipam: "true"
+  ```
+  * It's advisable to taint your windows nodes
+  * Any pods that target windows will need to have the have the following attributes set in their manifest
+  ```yaml
+    nodeSelector:
+      kubernetes.io/os: windows
+      kubernetes.io/arch: amd64
+  ```
+
+https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html
+<!-- markdownlint-disable -->
 ## Requirements
 
 | Name | Version |
@@ -255,8 +280,8 @@ Available targets:
 
 | Name | Version |
 |------|---------|
-| <a name="provider_aws"></a> [aws](#provider\_aws) | >= 3.56 |
-| <a name="provider_random"></a> [random](#provider\_random) | >= 2.0 |
+| <a name="provider_aws"></a> [aws](#provider\_aws) | 4.48.0 |
+| <a name="provider_random"></a> [random](#provider\_random) | 3.4.3 |
 
 ## Modules
 
@@ -295,7 +320,7 @@ Available targets:
 | <a name="input_after_cluster_joining_userdata"></a> [after\_cluster\_joining\_userdata](#input\_after\_cluster\_joining\_userdata) | Additional `bash` commands to execute on each worker node after joining the EKS cluster (after executing the `bootstrap.sh` script). For more info, see https://kubedex.com/90-days-of-aws-eks-in-production | `list(string)` | `[]` | no |
 | <a name="input_ami_image_id"></a> [ami\_image\_id](#input\_ami\_image\_id) | AMI to use. Ignored if `launch_template_id` is supplied. | `list(string)` | `[]` | no |
 | <a name="input_ami_release_version"></a> [ami\_release\_version](#input\_ami\_release\_version) | EKS AMI version to use, e.g. For AL2 "1.16.13-20200821" or for bottlerocket "1.2.0-ccf1b754" (no "v"). For AL2 and bottlerocket, it defaults to latest version for Kubernetes version. | `list(string)` | `[]` | no |
-| <a name="input_ami_type"></a> [ami\_type](#input\_ami\_type) | Type of Amazon Machine Image (AMI) associated with the EKS Node Group.<br>Defaults to `AL2_x86_64`. Valid values: `AL2_x86_64`, `AL2_x86_64_GPU`, `AL2_ARM_64`, `BOTTLEROCKET_x86_64`, and `BOTTLEROCKET_ARM_64`. | `string` | `"AL2_x86_64"` | no |
+| <a name="input_ami_type"></a> [ami\_type](#input\_ami\_type) | Type of Amazon Machine Image (AMI) associated with the EKS Node Group.<br>Defaults to `AL2_x86_64`. Valid values: `AL2_x86_64, AL2_x86_64_GPU, AL2_ARM_64,CUSTOM, BOTTLEROCKET_ARM_64, BOTTLEROCKET_x86_64, BOTTLEROCKET_ARM_64_NVIDIA, BOTTLEROCKET_x86_64_NVIDIA, WINDOWS_CORE_2019_x86_64, WINDOWS_FULL_2019_x86_64, WINDOWS_CORE_2022_x86_64, WINDOWS_FULL_2022_x86_64`. | `string` | `"AL2_x86_64"` | no |
 | <a name="input_associated_security_group_ids"></a> [associated\_security\_group\_ids](#input\_associated\_security\_group\_ids) | A list of IDs of Security Groups to associate the node group with, in addition to the EKS' created security group.<br>These security groups will not be modified. | `list(string)` | `[]` | no |
 | <a name="input_attributes"></a> [attributes](#input\_attributes) | ID element. Additional attributes (e.g. `workers` or `cluster`) to add to `id`,<br>in the order they appear in the list. New attributes are appended to the<br>end of the list. The elements of the list are joined by the `delimiter`<br>and treated as a single ID element. | `list(string)` | `[]` | no |
 | <a name="input_before_cluster_joining_userdata"></a> [before\_cluster\_joining\_userdata](#input\_before\_cluster\_joining\_userdata) | Additional `bash` commands to execute on each worker node before joining the EKS cluster (before executing the `bootstrap.sh` script). For more info, see https://kubedex.com/90-days-of-aws-eks-in-production | `list(string)` | `[]` | no |
@@ -465,7 +490,7 @@ In general, PRs are welcome. We follow the typical "fork-and-pull" Git workflow.
 
 ## Copyright
 
-Copyright © 2017-2022 [Cloud Posse, LLC](https://cpco.io/copyright)
+Copyright © 2017-2023 [Cloud Posse, LLC](https://cpco.io/copyright)
 
 
 
--- a/ami.tf
+++ b/ami.tf
@@ -8,7 +8,7 @@ locals {
     "BOTTLEROCKET_x86_64" : "x86_64",
     "BOTTLEROCKET_ARM_64" : "aarch64"
     "BOTTLEROCKET_ARM_64_NVIDIA" : "-gpu"
-    "BOTTLEROCKET_x86_64_NVIDIA " : "-gpu"
+    "BOTTLEROCKET_x86_64_NVIDIA" : "-gpu"
     "WINDOWS_CORE_2019_x86_64" : ""
     "WINDOWS_FULL_2019_x86_64" : ""
     "WINDOWS_CORE_2022_x86_64" : ""
--- a/ami.tf
+++ b/ami.tf
@@ -7,6 +7,12 @@ locals {
     "AL2_ARM_64" : "-arm64",
     "BOTTLEROCKET_x86_64" : "x86_64",
     "BOTTLEROCKET_ARM_64" : "aarch64"
+    "BOTTLEROCKET_ARM_64_NVIDIA" : "-gpu"
+    "BOTTLEROCKET_x86_64_NVIDIA " : "-gpu"
+    "WINDOWS_CORE_2019_x86_64" : ""
+    "WINDOWS_FULL_2019_x86_64" : ""
+    "WINDOWS_CORE_2022_x86_64" : ""
+    "WINDOWS_FULL_2022_x86_64" : ""
   }
 
   ami_kind = split("_", var.ami_type)[0]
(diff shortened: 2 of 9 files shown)


ChrisMcKee:
@osterman How often are PRs reviewed ooi?

iam.tf
```diff
@@ -51,9 +51,9 @@ resource "aws_iam_role_policy_attachment" "existing_policies_for_eks_workers_rol
 # Create a CNI policy that is a merger of AmazonEKS_CNI_Policy and required IPv6 permissions
 # https://github.com/SummitRoute/aws_managed_policies/blob/master/policies/AmazonEKS_CNI_Policy
 # https://docs.aws.amazon.com/eks/latest/userguide/cni-iam-role.html#cni-iam-role-create-ipv6-policy
-
+# https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html
 data "aws_iam_policy_document" "ipv6_eks_cni_policy" {
-  count = local.create_role && var.node_role_cni_policy_enabled ? 1 : 0
+  count = local.create_role && var.node_role_cni_policy_enabled || local.create_role && can(regex("WINDOWS", var.ami_type)) ? 1 : 0
```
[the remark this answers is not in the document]
ChrisMcKee: I'm checking this one.
To run windows you have to run linux nodes as well to provide the funtions "currently" not supported on windows by k8s (2023 is supposed to be a 'focus on windows' year so who knows)
The changes were based on our working cluster. 
The minimum requirements are the policy on the cluster (hence my pr [on that](https://github.com/cloudposse/terraform-aws-eks-cluster/pull/175), and the policy and taints on windows nodes and a config map for IPAM.
The policy is to allow assignment of IPs and I _think_ I set it to use it on windows so that would happen.
I need to retest as we dont use the built in role and I havent looked at this since January 😉

ChrisMcKee:
Just trying to work out why it's stopped working. It all spins up but the nodes don't link to eks..
I'm using the completeexample though modified to test this too. I'll update when it plays ball 🤔

nitrocode:
@ChrisMcKee resolved conflicts and merged changes in

ChrisMcKee:
Relies on https://github.com/cloudposse/terraform-aws-eks-cluster/pull/175

ChrisMcKee:
Probably made git cry but rebased off master and squashed out all the autoformat commits etc.
I've pulled the windows example out of examples.tf
I'll add one in a separate PR as there's a bit of faff required when you use a userscript and an ami is used as it needs to append the node to the aws_auth config map (like usual) but the windows node requires an extra permission for ipam.

ChrisMcKee:
@aknysh all good now; I've pushed out a few windows node groups to our internal clusters using this combined with the cluster pr.

ChrisMcKee:
@aknysh sorry for the repeated changes after the fact; I hit an issue randomly after using my fork since the pr went up and, as you do, I assumed I'd missed a commit or messed up. Turned out to be a flipping bug with kube-proxy 🤦 The module as it stands works.

The only remaining issue I have with it is the validation in the variables file over release; the regex is obviously setup like that for a reason but the Windows AMI's dont follow the same setup
being more like `1.24-2023.02.14`

README.md
```diff
@@ -243,19 +243,51 @@ Available targets:
 ```
 <!-- markdownlint-restore -->
 <!-- markdownlint-disable -->
+## Windows Managed Node groups
```
nitrocode, line 246: If this is manually added, it will be overwritten. You have to add it to README.yaml

docs/windows.md
```diff
@@ -0,0 +1,32 @@
+<!-- markdownlint-disable -->
+## Windows Managed Node groups
+  Windows managed node-groups have a few pre-requisites.
```
nitrocode, line 3: Can we remove the prefixed space here ?

examples/complete/main.tf
```diff
@@ -3,8 +3,8 @@ provider "aws" {
 }
 
 module "label" {
-  source  = "cloudposse/label/null"
-  version = "0.25.0"
+  source = "cloudposse/label/null"
+  #version = "0.25.0"
```
nitrocode, line 7: We'll need to keep this version

examples/complete/main.tf
```diff
@@ -135,7 +133,7 @@ module "eks_cluster" {
   context = module.this.context
 }
 
-module "eks_node_group" {
+module "eks_nix_node_group" {
```
nitrocode, line 136: Unsure why this is modified

examples/complete/main.tf
```diff
@@ -147,19 +145,19 @@ module "eks_node_group" {
...
   update_config                 = var.update_config
 
-  after_cluster_joining_userdata = var.after_cluster_joining_userdata
+  //after_cluster_joining_userdata = var.after_cluster_joining_userdata
 
   ami_type            = var.ami_type
   ami_release_version = var.ami_release_version
 
-  before_cluster_joining_userdata = [var.before_cluster_joining_userdata]
+  //before_cluster_joining_userdata = [var.before_cluster_joining_userdata]
```
nitrocode, line 160: Please revert all these example changes

iam.tf
```diff
@@ -51,9 +51,9 @@ resource "aws_iam_role_policy_attachment" "existing_policies_for_eks_workers_rol
 # Create a CNI policy that is a merger of AmazonEKS_CNI_Policy and required IPv6 permissions
 # https://github.com/SummitRoute/aws_managed_policies/blob/master/policies/AmazonEKS_CNI_Policy
 # https://docs.aws.amazon.com/eks/latest/userguide/cni-iam-role.html#cni-iam-role-create-ipv6-policy
-
+# https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html
 data "aws_iam_policy_document" "ipv6_eks_cni_policy" {
-  count = local.create_role && var.node_role_cni_policy_enabled ? 1 : 0
+  count = local.create_role && (var.node_role_cni_policy_enabled) ? 1 : 0
```
nitrocode, line 56: revert

ami.tf
```diff
@@ -7,9 +7,15 @@ locals {
     "AL2_ARM_64" : "-arm64",
     "BOTTLEROCKET_x86_64" : "x86_64",
     "BOTTLEROCKET_ARM_64" : "aarch64"
+    "BOTTLEROCKET_ARM_64_NVIDIA" : "-gpu"
```
nitrocode, line 10: do you mind converting this to a normal hcl map?

```suggestion
    BOTTLEROCKET_ARM_64_NVIDIA = "-gpu"
```

versions.tf
```diff
@@ -1,12 +1,10 @@
 terraform {
-  required_version = ">= 0.14.11"
+  required_version = ">= 0.15.00"
```
nitrocode, line 2: Why is this bumped?

versions.tf
```diff
@@ -1,12 +1,10 @@
...
 
   required_providers {
     aws = {
-      source = "hashicorp/aws"
-      # retrieve launch template by ID starts at 3.21.0
-      # update_config starts at 3.56
-      version = ">= 3.56"
+      source  = "hashicorp/aws"
+      version = ">= 4.0"
```
nitrocode, line 7: Why is this bumped?

README.md
nitrocode: oh I see what you did with `windows.md`

docs/windows.md
ChrisMcKee: Yeah; no idea where that came from

examples/complete/main.tf
ChrisMcKee: I added an eks_nt_node_group so it was separating the groups; I was going to ask what the opinion was. could be a separate example instead. As the cluster change is necessary and I was reverting this to be 99% back to its original test sample I figured I'd lower the threshold on this pass.

examples/complete/main.tf
ChrisMcKee: 👍

versions.tf
ChrisMcKee: That versions probably overly loose as the underlying api-sdk only supports windows from the December release https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md#4480-december-19-2022
So 4.48.0

versions.tf
ChrisMcKee: Presumption that the world should move on 😆 
Honestly 15 was when most stuff became stableish and I couldn't find a simple "AWS Provider 4 supports version x above" in the repo'; then I remembered its burried in the use-provider option 🙄 
image

versions.tf
```diff
@@ -3,10 +3,8 @@ terraform {
 
   required_providers {
     aws = {
-      source = "hashicorp/aws"
-      # retrieve launch template by ID starts at 3.21.0
-      # update_config starts at 3.56
-      version = ">= 3.56"
+      source  = "hashicorp/aws"
+      version = ">= 4.48"
```
nitrocode, line 7: Like before, please add a comment on the exact reason for the bump

examples/complete/main.tf
nitrocode: I'm not sure i understand why the module name has changed tho. Could we change it back?

ChrisMcKee: Because the windows module was eks_nt_node_group. 
If its that much of an issue I'll revert it when I get out of this bike shed 😉

ChrisMcKee: All done

ami.tf
ChrisMcKee: I updated the lot

ChrisMcKee:
@nitrocode @aknysh This will also resolve #137

main.tf
```diff
@@ -44,6 +44,12 @@ locals {
     }
   )
   node_group_tags = merge(local.node_tags, local.autoscaler_enabled ? local.autoscaler_tags : null)
+
+  windows_taint = [{
+    key    = "OS"
+    value  = "Windows"
+    effect = "NO_SCHEDULE"
+  }]
```
Nuru, line 52: A taint like this, with a non-standard key, should at least be optional. I would argue it should be supplied by the user if desired, because they may already be using a different key for tainting by OS or may not want to taint at all and instead rely on node selectors with the ["well-known"](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-os) `kubernetes.io/os` label.

main.tf
```diff
@@ -71,11 +78,12 @@ locals {
...
-    ami_type        = local.launch_template_ami == "" ? var.ami_type : null
-    capacity_type   = var.capacity_type
-    labels          = var.kubernetes_labels == null ? {} : var.kubernetes_labels
-    taints          = var.kubernetes_taints
+    instance_types = sort(var.instance_types)
+    ami_type       = local.launch_template_ami == "" ? var.ami_type : null
+    capacity_type  = var.capacity_type
+    labels         = var.kubernetes_labels == null ? {} : var.kubernetes_labels
+
+    taints          = local.is_windows ? concat(local.windows_taint, var.kubernetes_taints) : var.kubernetes_taints
```
Nuru, line 86: This taint should be optional and arguably default to not set. See comment above where the taint is defined.

userdata.tpl
```diff
@@ -4,6 +4,7 @@ Content-Type: multipart/mixed; boundary="/:/+++"
 --/:/+++
 Content-Type: text/x-shellscript; charset="us-ascii"
 #!/bin/bash
+set -ex
```
Nuru, line 7: I am not a fan of this in userdata scripts because the echo can reveal secrets (echoed data gets logged, logs get shipped to a log aggregator, people without any access to the machines get to read the logs via the log aggregator) and given that users can add snippets to this script without the benefit of context, this risks blowing up a whole node group by something as simple as `grep` failing to find the pattern in the file.

Nuru (commented):
I know this was already merged, but I made some comments I hope will be addressed in the future.

main.tf
ChrisMcKee: You'd use the nodeselector but the taint stops the mass of ingress/nginx/haproxy/grafana helm default installs from pointlessly trying to install on the nodes. 
I did leave it out originally and just had it in the docs. 🤷‍♀️

userdata.tpl
ChrisMcKee: Alternatively the userscript could be surrounded in the same way as the powershell userscript surrounds the pre/post. 
-e just ensures that an error raises an error code
-x write to standard error / expand commmand before execution. 

Witout it the shipped log would be quite useless / we don't ship userdata but I expect ymmv.

Nuru: I would leave the `set -ex` up to the user to set. `-x` can expose secrets the script accesses, `-e` doesn't just ensure the error raises an error code, is stops the execution of the script if a command returns a non-zero status. Both `grep` and `diff` routinely return non-zero status in normal operation.

Issue #135, opened 2022-12-16
Instance type order preference not respected


xeivieni:
Found a bug? Maybe our [Slack Community](https://slack.cloudposse.com) can help. 

[![Slack Community](https://slack.cloudposse.com/badge.svg)](https://slack.cloudposse.com)

## Describe the Bug
As stated in the AWS node group documentation here https://docs.aws.amazon.com/eks/latest/userguide/managed-node-groups.html 
> Managed node groups use the order of instance types passed in the API to determine which instance type to use first when fulfilling On-Demand capacity. For example, you might specify three instance types in the following order: c5.large, c4.large, and c3.large. When your On-Demand Instances are launched, the managed node group fulfills On-Demand capacity by starting with c5.large, then c4.large, and then c3.large

However, the module overrides the order of the instance_types list by doing a sort on it https://github.com/cloudposse/terraform-aws-eks-node-group/blob/master/main.tf#L74

## Expected Behavior
The order of preference of the instance_types shall be passed to the node_group resource as specified by the user

## Steps to Reproduce
1. Create a node group with three instance types
2. Apply 
3. Change the order of the instance types
4. plan
5. See no changes

Nuru:
closed by #142

PR #142, opened 2023-03-14
Do not sort instance types

xeivieni:
## what

Remove sorting on instance type list in the node group definition

## why

Because the order of the list is used to define priorities on the type of instance to use.

## references

- https://github.com/cloudposse/terraform-aws-eks-node-group/issues/135
- https://docs.aws.amazon.com/eks/latest/userguide/managed-node-groups.html

> Managed node groups use the order of instance types passed in the API to determine which instance type to use first when fulfilling On-Demand capacity. For example, you might specify three instance types in the following order: c5.large, c4.large, and c3.large. When your On-Demand Instances are launched, the managed node group fulfills On-Demand capacity by starting with c5.large, then c4.large, and then c3.large


Gowiem:
This looks good to me, but since @Nuru is the defacto lead on this module, I'm going to pull him in. Particularly because he is the one who approved #54 where this was originally introduced. 

@Nuru any strong opinions here before we move it forward?

Nuru:
@xeivieni Thank you for this PR.

The `random_pet`'s `keepers` should match the behavior of the Terraform provider. When inputs change that cause Terraform to want to replace the node group (rather than update it in place), a new `random_pet` should be generated. Some things have changed since this module was first written, so please re-check:

- Does changing the order of the on-demand instance types cause a new node group to be created?
- Does changing the set of on-demand instance types cause a new node group to be created?

The `keepers` should be set to mimic those answers. My guess is that the current 

```hcl
instance_types = join(",", sort(local.ng.instance_types))
```

is probably not the best choice.

main.tf
```diff
@@ -106,7 +105,7 @@ resource "random_pet" "cbd" {
   keepers = {
     node_role_arn  = local.ng.node_role_arn
     subnet_ids     = join(",", local.ng.subnet_ids)
-    instance_types = join(",", local.ng.instance_types)
+    instance_types = join(",", sort(local.ng.instance_types))
```
Nuru, line 108: Confirmed that changing the instance types forces a new node group:
```suggestion
    instance_types = join(",", local.ng.instance_types)
```

Nuru (approved):

PR #149, opened 2023-07-04
add core_count and threads_per_core options to launch templates

Dmitry1987:
## what

added core_count and threads_per_core options in order to run instances with no hyperthreading, for applications that need maximize single core performance (in some cases it's required). 

## why
the config option is available in the module but was not available in variables

## references
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#example-usage


Dmitry1987:
this is good stuff, why nobody wants to approve? 😅 
for single threaded apps, very much required functionality actually...

PR #150, opened 2023-08-03
Better block device support

Nuru:
## Notable Changes

- With this PR/release, the default type of block device changes from `gp2` to `gp3`. If you were relying on the default, this will cause your node group to change, but it should be without interruption. 
- We no longer automatically apply a custom taint to Window nodes. Pods should select (or not) Windows nodes via the `kubernetes.io/os` tag. If that is not sufficient, you are free to add your own "NO_EXECUTE" taint via `kubernetes_taints`

## what

- Better support for block device mapping
- Update dependency `terraform-aws-security-group` to current v2.2.0
- Revert portions of #139

## why

- Take advantage of `optional()` to allow the block device mapping input to be fully specified, with defaults, rather than the previous `list(any)`, which had no type checking and did not advertise which features were or were not supported
- Closes #134 
- Bad practices that were not caught in time


max-lobur (approved):

variables.tf
```diff
@@ -160,21 +164,27 @@ variable "capacity_type" {
...
     EOT
   # See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#ebs
-  default = [{
-    device_name           = "/dev/xvda"
-    volume_size           = 20
-    volume_type           = "gp2"
-    encrypted             = true
-    delete_on_termination = true
-  }]
+  default = { "/dev/xvda" = { ebs = {} } }
```
max-lobur, line 187: Suggestion: Add known operating systems and their default storage in description notes

PR #151, opened 2023-09-26
Add `force_update_version` and `replace_node_group_on_version_update` variables

aknysh:
## what
* Add `force_update_version` and `replace_node_group_on_version_update` variables

## why
* `force_update_version` allows forcing version update if existing pods are unable to be drained due to a pod disruption budget issue. Default is `false` for backwards compatibility

* If the variable `replace_node_group_on_version_update` is set to `true` and the EKS cluster is updated to a new Kubernetes version, the Node Groups will be replaced instead of updated in-place. This is useful when updating very old EKS clusters to a new Kubernetes version where some old releases prevent nodes from being drained (due to PodDisruptionBudget or taint/toleration issues), but replacing the Node Groups works without forcing the pods to leave the old nodes by using the `force_update_version` variable. This is related, for example, to `istio`. Default is `false` for backwards compatibility

## references
* https://docs.aws.amazon.com/eks/latest/userguide/managed-node-update-behavior.html
* https://github.com/aws/containers-roadmap/issues/1282


kcarter-bread (approved):
Approved

variables.tf
```diff
@@ -450,3 +450,15 @@ variable "detailed_monitoring_enabled" {
   default     = false
   description = "The launched EC2 instance will have detailed monitoring enabled. Defaults to false"
 }
+
+variable "force_update_version" {
+  type        = bool
+  default     = false
+  description = "Force Kubernetes version update if existing Pods are unable to be drained from the nodes due to PodDisruptionBudget or taint/toleration issues"
```
Nuru, line 457:
```suggestion
  description = "When updating the Kubernetes version, force Pods to be removed even if PodDisruptionBudget or taint/toleration issues would otherwise prevent them from being removed (and cause the update to fail)"
```

variables.tf
```diff
@@ -450,3 +450,15 @@ variable "detailed_monitoring_enabled" {
...
+}
+
+variable "replace_node_group_on_version_update" {
+  type        = bool
+  default     = false
+  description = "Force Node Group replacement when updating to a new Kubernetes version. If set to `true`, the Node Groups will be replaced instead of updated in-place"
```
Nuru, line 463:
```suggestion
  description = "Force Node Group replacement when updating to a new Kubernetes version. If set to `false` (the default), the Node Groups will be updated in-place."
```

Nuru (changes requested):

Nuru (approved):

PR #178, opened 2024-05-10
Test

### LINKED ISSUE #123: after_cluster_joining_userdata only allows one element

Darsh8790:


Nuru:
@Darsh8790 Thank you for this PR. However, as it is, it is not acceptable.

1. Do not change default values. That would be a breaking change, which we try to avoid.
2. I believe your regex change would break AL2 support. If it does not, please explain why in comments.
3. If we are going to add support for AL2023, then we want to add support for both `x86_64` and `arm64` architectures.
4. After making changes but before your final commit and push, please run:

```bash
make init
make precommit/terraform # requires `docker run` be available
```

5. Please fill in the required PR comment fields.

Nuru:
@Darsh8790 Note that you will need to provide backward compatibility for AL2 users, so you cannot just replace the existing `userdata.tpl` with one for AL2023. Probably you should use the same inputs and have different templates for AL2 and AL2023, like the way we have different templates for AL2 and Windows.

Nuru:
## Reviewers please note

- Extra configuration for AL2 is done by passing arguments to [bootstrap.sh](https://github.com/awslabs/amazon-eks-ami/blob/25b6f34595f45080e013fb5e82761269f887795f/templates/al2/runtime/bootstrap.sh). 
- For AL2023, configuration is handled via [nodeadm](https://awslabs.github.io/amazon-eks-ami/nodeadm/).

In order to support AL2023, it is not only the AMI selection that needs to be updated, but also everything configured via `userdata`.

PR #179, opened 2024-05-10
Remove ssh-egress security group rule

m-wynn:
## what

* Delete the unnecessary "ssh-egress" security-group rule.  It does not provide anything related to SSH.

⚠️  This might break environments for people who are accidentally relying on the wide security group rules for non-ssh outbound access.  However it will allow others to be more secure and tighten down their security posture.

## why

AWS Security Groups are stateful, meaning that any inbound traffic that is allowed will automatically permit the return outbound traffic, regardless of outbound rules.  Security Groups track TCP connections in order to determine which packets are responses to allowed inbound traffic.

When you establish an SSH connection, you're initiating an inbound connection to the server. The server's response to your connection, as well as any subsequent communication, is considered outbound traffic from the perspective of the security group.

Because of the stateful nature of AWS Security Groups, once you've allowed inbound SSH traffic, the return traffic (outbound from the server) is automatically allowed, even if your outbound rules don't explicitly permit it.

Therefore, an outbound rule that allows traffic on all ports is not necessary for the operation of inbound SSH.

## references

* https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/security-group-connection-tracking.html

### DIFF
--- a/security-group.tf
+++ b/security-group.tf
@@ -27,14 +27,6 @@ module "ssh_access" {
       from_port   = 22
       to_port     = 22
       description = "Allow SSH ingress"
-      },
-      {
-        key         = "ssh-egress"
-        type        = "egress"
-        from_port   = 0
-        to_port     = 65535
-        protocol    = "tcp"
-        description = "Allow SSH egress"
     }]
   }]
 


Nuru (changes requested):
@m-wynn Thank you for this PR. 

When you say:

> An outbound rule that allows traffic on all ports is not necessary for the operation of inbound SSH.

I believe you are correct. However, I'm not sure that users are using it that way. When the source is a security group (e.g. a bastion host), they may be intending to allow a user from that SG to run other commands on the node group that access that SG, such as `curl` or `ftp`. So, although this is indeed tightening a loophole, it nevertheless is a breaking change. Would you mind explaining this in your PR comments, and explaining how to restore unlimited egress to the source SG or CIDR if that is desired? 

(Reverting will be a lot easier if you make the egress rule optional with a new variable `ssh_allow_all_egress`. I propose you make that a boolean variable that defaults to `null`. When `null`, all egress is allowed when `ssh_access_security_group_ids` is supplied but not allowed if it is not, and the source CIDR is wide open. How does that sound to you?)


Also, we don't have a test for SSH access. Would you be willing to add one? (It requires you know `go` and Terratest, and is a non-trivial amount of work, so I understand if you feel you can't or don't want to do it.)

Nuru:
@m-wynn I'm closing this as "wontfix" for the reasons state above. In short, a key reason to get a shell on a node via SSH is to be able to use it as a bastion host to access other hosts in the VPC. And in general most nodes already have unlimited egress. 

I'd be open to a PR that provides more fine-grained control over the security group, but I don't want a change like this pushed on people.

PR #180, opened 2024-05-20
Made necessary changes for worker node AL2023 in version 1.0.1

### LINKED ISSUE #123: after_cluster_joining_userdata only allows one element

Darsh8790:
## what

 Added changes for AL2023 worker node AMI based on version tag v1.0.0.

## why

     Currently, there are no changes for AL2023, even in the latest tag 2.12.0. The necessary modifications for AL2023 EKS worker nodes are not present.
    The existing EKS nodes running Amazon Linux 2 have multiple vulnerabilities. To remediate these issues, we need to upgrade our EKS worker nodes to AL2023 worker nodes.

## references

     We have tested these changes in our infrastructure, and they are working well.
    We request that these changes be merged and released as a new tag, 1.0.1, which will include all the updates up to tag 1.0.0, along with the changes for AL2023 worker nodes.

### DIFF
--- a/ami.tf
+++ b/ami.tf
@@ -5,6 +5,7 @@ locals {
     "AL2_x86_64" : "",
     "AL2_x86_64_GPU" : "-gpu",
     "AL2_ARM_64" : "-arm64",
+    "AL2023_x86_64_standard" : "x86_64"
   }
 
   # Kubernetes version priority (first one to be set wins)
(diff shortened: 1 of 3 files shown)


Nuru:
@Darsh8790 Thank you for this PR. 

I appreciate how much work you put into this. However, we cannot accept it as-is. If you do not want to work on it further, I understand, and we will use this to guide our future support for AL2023.

The main issue I see right off the bat is that this PR effectively breaks support for any AMI family other than AL2023. We would want to maintain backward compatibility for any AMIs Amazon continues to support.

Also, for this particular module, we are not supporting changes to v1. Although in general we want to allow for modifications for old major versions, this module's v1 was never really solid enough for that. So we would want these modifications to be made to the current version, which as of this writing is v2.12.0. (Note that even if we made the changes to v1, we would want to maintain support for older AMIs.)

Let us know how you'd like to proceed.

QuentinBtd:
Working on https://github.com/cloudposse/terraform-aws-eks-node-group/pull/185 to add support of AL2023 with latest module version :)

PR #186, opened 2024-06-10
Add support for AL2023

### LINKED ISSUE #155: Unflattened output for resources
### Describe the Bug

The output value `eks_node_group_resources` does not flatten out the array produced from the `[*]` reference (or just reference the zero element since the `create_before_destroy` toggle is already used as a condition). The result is that the output of resources is a 2 dimensional array.

### Expected Behavior

`eks_node_group_resources` should be a single dimensional array of resources matching the underlying `aws_eks_node_group` module output.

### Steps to Reproduce

Create a node group with the module and observe the output.

### Screenshots

_No response_

### Environment

_No response_

### Additional Context

_No response_

### LINKED ISSUE #174: Unavailable AMI Type Selection: Al2023
### Describe the Feature

The AMI type "Al2023" (specifically, "amazon-eks-node-al2023-x86_64-standard-1.29.0-20240415") is not available for selection at the moment.

### Expected Behavior

We should be able to use Al2023 AMI type for worker nodes.

### Use Case

We should be able to use Al2023 AMI type for worker nodes.

### Describe Ideal Solution

We should be able to use Al2023 AMI type for worker nodes.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### LINKED ISSUE #183: Support multiple lines in before_cluster_joining_userdata and after_cluster_joining_userdata?
### Describe the Bug

When I specify multiple items in the before_cluster_joining_userdata list, I get an error on plan:
```
│ You may not specify more than one `before_cluster_joining_userdata`.
```

There's actually a test [here](https://github.com/cloudposse/terraform-aws-eks-node-group/blob/main/variables.tf#L323) which limits this list to one item!!
```terraform
variable "before_cluster_joining_userdata" {
  type        = list(string)
  default     = []
  description = "Additional `bash` commands to execute on each worker node before joining the EKS cluster (before executing the `bootstrap.sh` script). For more info, see https://kubedex.com/90-days-of-aws-eks-in-production"
  validation {
    condition = (
      length(var.before_cluster_joining_userdata) < 2
    )
    error_message = "You may not specify more than one `before_cluster_joining_userdata`."
  }
}
```

Can we support a list, or update the documentation on this and after_cluster_joining_userdata?

### Expected Behavior

The docs say:
> before_cluster_joining_userdata list(string)
> Description: Additional `bash` commands to execute on each worker node before joining the EKS cluster (before executing > the `bootstrap.sh` script). For more info, see https://kubedex.com/90-days-of-aws-eks-in-production
> Default: []

### Steps to Reproduce

```
module "nodegroups" {
  source = "git::https://github.com/cloudposse/terraform-aws-eks-node-group.git//?ref=6ccf1a9e30101540d479bbc9e185513476fff17a"
  ...
  before_cluster_joining_userdata = [
    "echo one",
    "echo two",
   ...

### Screenshots

_No response_

### Environment

_No response_

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by Nuru**
## New Features, Breaking Changes

tl;dr Upgrading to this version will likely cause your node group to be replaced, but otherwise should not have much impact for most users.

The major new feature in this release is support for Amazon Linux 2023 (AL2023). EKS support for AL2023 is still evolving, and this module will evolve along with that. Some detailed configuration options (e.g. KubeletConfiguration JSON) are not yet supported, but the basic features are there.

The other big improvements are in immediately applying changes and in selecting AMIs, as explained below. 

Along with that, we have dropped some outdated support and changed the `eks_node_group_resources` output, resulting in minor breaking changes that we expect do not affect many users.

### Create Before Destroy is Now the Default

Previously, when changes forced the creation of a new node group, the default behavior for this module was to delete the existing node group and then create a replacement. This is the default for Terraform, motivated in part by the fact that the node group's name must be unique, so you cannot create the new node group with the same name as the old one while the old one still exists.

With version 2 of this module, we recommended setting `create_before_destroy` to `true` to enable this module to create a new node group (with a partially randomized name) before deleting the old one, allowing the new one to take over for the old one. For backward compatibility, and because changing this setting always results in creating a new node group, the default setting was set to `false`.

With this release, the default setting of `create_before_destroy` is now `true`, meaning that if left unset, any changes requiring a new node group will cause a new node group to be created first, and then the existing node group to be deleted. If you have large node groups or small quotas, this can fail due to having the 2 node groups running at the same time. 

#### Random name length now configurable

In order to support "create before destroy" behavior, this module uses the `random_pet`
resource to generate a unique pet name for the node group, since the node group name
must be unique, meaning the new node group must have a different name than not only the old one, but also all other node groups you have. Previously, the "random" pet name was 1 of 452 possible names, which may not be enough to avoid collisions when using a large number of node groups. 

To address this, this release introduces a new variable, `random_pet_length`, that controls the number of pet names concatenated to form the random part of the name. The default remains 1, but now you can increase it if needed. Note that changing this value will always cause the node group name to change and therefore the node group to be replaced.

### Immediately Apply Launch Template Changes

This module always uses a launch template for the node group. If one is not supplied, it will be created. 

In many cases, changes to the launch template are not immediately applied by EKS. Instead, they only apply to Nodes launched after the template is changed. Depending on other factors, this may mean weeks or months pass before the changes are actually applied. 

This release introduces a new variable, `immediately_apply_lt_changes`, to address this. When set to true, any changes to the launch template will cause the node group to be replaced, ensuring that all the changes are made immediately. (Note: you may want to adjust the `node_group_terraform_timeouts` if you have big node groups.)

The default value for `immediately_apply_lt_changes` is whatever the value of `create_before_destroy` is.

### Changes in AMI selection

Previously, if the created launch template needed to supply an AMI ID (which is only the case if you supplied kubelet or bootstrap options), unless you specified a specific AMI ID, this module picked the "newest" AMI that met the selection criteria, which in turn was based on the AMI Name. The problem with that was that the "newest" might not be the latest Kubernetes version. It might be an older version that was patched more recently, or simply finished building a little later than the latest version.

Now that AWS explicitly [publishes](https://docs.aws.amazon.com/eks/latest/userguide/retrieve-ami-id.html) the AMI ID corresponding to the latest (or, more accurately, "recommended") version of their AMIs via SSM Public Parameters, the module uses that instead. This is more reliable and should eliminate the version regression issues that occasionally happened before.

#### The `ami_release_version` input has been updated

The `ami_release_version` input has been updated. It is the value that you can supply to [`aws_eks_node_group`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group) to track a specific patch version of Kubernetes. The previous validation for this variable was incorrect.  

- For Amazon Linux, it is the "Release version" from [Amazon AMI Releases](https://github.com/awslabs/amazon-eks-ami/releases)
- For Bottlerocket, it is the release tag from [Bottlerocket Releases](https://github.com/bottlerocket-os/bottlerocket/releases) without the "v" prefix.
- For Windows, it is "AMI version" from [AWS docs](https://docs.aws.amazon.com/eks/latest/userguide/eks-ami-versions-windows.html).

Note that unlike AMI names, release versions never include the "v" prefix.

Examples of AMI release versions based on OS:

- Amazon Linux 2 or 2023: 1.29.3-20240531
- Bottlerocket: 1.18.0 or 1.18.0-7452c37e # note commit hash prefix is 8 characters, not GitHub's default 7
- Windows: 1.29-2024.04.09

#### Unsupported `userdata` now throws an error

Node configuration via `userdata` is different for each OS. This module has 4 inputs related to Node configuration that end up using `userdata`:

  1. `before_cluster_joining_userdata`
  2. `kubelet_additional_options`
  3. `bootstrap_additional_options`
  4. `after_cluster_joining_userdata`

but they do not all work for all OSes, and none work for Botterocket. Previously, they were silently ignored in some cases. Now they throw an error when set for an unsupported OS. 

Note that for all OSes, you can bypass all these inputs and supply your own fully-formed, base64 encoded `userdata` via `userdata_override_base64`, and this module will pass it along unmodified.

#### Multiple lines supported in `userdata` scripts

All the `userdata` inputs take lists, because they are optional inputs. Previously, lists were limited to single elements. Now the list can be any length, and the elements will be combined.

### Kubernetes Version No Longer Inferred from AMI

Previously, if you specified an AMI ID, the Kubernetes version would be deduced from the AMI ID name. That is not sustainable as new OSes are launched, so the module no longer tries to do that. If you do not supply the Kubernetes version, the EKS cluster's Kubernetes version will be used. 

### Output `eks_node_group_resources` changed

The `aws_eks_node_group.resources` attribute is a "list of objects containing information about underlying resources." Previously, this was output via `eks_node_group_resources` as a list of lists, due to a quirk of Terraform. It is now output as a list of resources, in order to align with the other outputs.

### Special Support for Kubernetes Cluster Autoscaler removed

This module used to takes some steps (mostly labeling) to try to help the Kubernetes Cluster Autoscaler. As the Cluster Autoscaler and EKS native support for it evolved, the steps taken became either [redundant](https://docs.aws.amazon.com/eks/latest/userguide/managed-node-groups.html) or [ineffective](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1558#issuecomment-909329290), so they have been dropped. 

- `cluster_autoscaler_enabled` has been deprecated. If you set it, you will get a warning in the output, but otherwise it has no effect.

### AWS Provider v5.8 or later now required

Previously, this module worked with AWS Provider v4, but no longer. Now v5.8 or later is required.

## Special Thanks

This PR builds on the work of @Darsh8790 (#178 and #180) and @QuentinBtd (#182 and #185). Thank you to both for your contributions.

---

---

## what

- Add initial support for EKS under Amazon Linux 2023 (AL2023)
- Improve AMI selection process
- Deprecate Kubernetes Cluster Autoscaler support

## why

- Amazon Linux 2023 (AL2023) is the latest offering from Amazon
- Previously, AMIs were selected by name and date, which occasionally led to undesirable results
- The support was either redundant or ineffective

## references

Documentation:
- [Cannot add tags to EKS Managed Node Group's ASG](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1558#issuecomment-909329290) 
- [EKS automatically tags nodes for Kubernetes Autoscaler](https://docs.aws.amazon.com/eks/latest/userguide/managed-node-groups.html)

Issues and Other PRs:

- Closes #155 
- Closes #174 
- Supersedes and closes #178 
- Supersedes and closes #180
- Supersedes and closes #182 
- Closes #183
- Supersedes and closes #185

### CODE REVIEW

**Review by QuentinBtd — APPROVED** [review]
Seems good for me! 🎸 

> This module used to takes some steps (mostly labeling) to try to help the Kubernetes Cluster Autoscaler. As the Cluster Autoscaler and EKS native support for it evolved, the steps taken became either redundant or ineffective (or were always ineffective), so they have been dropped.

I don't understand. Reading the [Cluster Autoscaler documentation](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md#auto-discovery-setup), tags are always used for auto discovery. 

Have I missed something?

**Comment by QuentinBtd** [comment]
Just tested:

I created a new node group with `ami_specifier` set to `amazon-eks-node-al2023-x86_64-standard-1.30-v20240531`. 
Then I changed the value to `amazon-eks-node-al2023-x86_64-standard-1.30-v20240605`, run an `apply` but the node group keep using the AMI corresponding to `amazon-eks-node-al2023-x86_64-standard-1.30-v20240531`

Only the `ami_ids` output changed.

**Comment by Nuru** [comment]
> Seems good for me! 🎸
> 
> But;
> 
> > This module used to takes some steps (mostly labeling) to try to help the Kubernetes Cluster Autoscaler. As the Cluster Autoscaler and EKS native support for it evolved, the steps taken became either redundant or ineffective (or were always ineffective), so they have been dropped.
> 
> I don't understand. Reading the [Cluster Autoscaler documentation](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md#auto-discovery-setup), tags are always used for auto discovery.
> 
> Have I missed something?

Yes. 😊 See [this issue](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1558#issuecomment-909329290) about why what the module was doing was ineffective and [this documentation](https://docs.aws.amazon.com/eks/latest/userguide/managed-node-groups.html) which say in relevant part:

> Amazon EKS tags managed node group resources so that they are configured to use the Kubernetes [Cluster Autoscaler](https://docs.aws.amazon.com/eks/latest/userguide/autoscaling.html).

**Comment by QuentinBtd** [comment]
> Yes. 😊 See [this issue](https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1558#issuecomment-909329290) about why what the module was doing was ineffective and [this documentation](https://docs.aws.amazon.com/eks/latest/userguide/managed-node-groups.html) which say in relevant part:
> 
> > Amazon EKS tags managed node group resources so that they are configured to use the Kubernetes [Cluster Autoscaler](https://docs.aws.amazon.com/eks/latest/userguide/autoscaling.html).

Thank you! I'm going to be able to clean up TF code of my dozens of node groups 😅

File: variables.tf [hunk 1]
```diff
@@ -125,43 +150,68 @@ variable "ami_type" {
...
 }
 
+variable "ami_specifier" {
+  type        = string
+  description = <<-EOT
+    OS-dependent specifier for one of the several AMIs that match OS, architecture, and Kubernetes 1.xx version.
+    If not specified the recommended/latest AMI for the given Kubernetes version will be used.
+    Unfortunately, the format of this value varies by OS, and we have not found documentation for it.
+    You can generally figure it out from the AMI name or description, and validate it by trying to retrieve
+    the SSM Public Parameter for the AMI ID.
```
QuentinBtd, line 185: AMI names are available here https://github.com/awslabs/amazon-eks-ami/releases

QuentinBtd:
Instances created with a launch template and whose AMI is specified in the template do not have kubelet/nodeadm configuration, so do not join the cluster

Nuru:
AL2023 requires that NodeConfig be supplied when AMI is configured in launch template.

Kubernetes has deprecated Kubelet command-line args in favor of a config file.

I am putting this on pause (help requested), to update our `userdata` handling as follows:

- Support, at a minimum, kube-reserved system-reserved eviction-hard eviction-soft settings for Kubelet for all OSes. This probably means an input with those elements so that they can be configured through the appropriate different mechanisms.
- Enable a way to provide a full KubeConfig file for all OSes without requiring a full userdata input.


References:

- https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html#launch-template-user-data
- https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html#mng-ami-id-conditions
- https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html#mng-ami-id-conditions

QuentinBtd (approved):
Good job 👍 💪

Benbentwo (approved):
LGTM

PR #191, opened 2024-08-08
feat: Add missed tags

MaxymVlasov:
## what

Add tags to resources where they missed


MaxymVlasov:
@aknysh fixed

MaxymVlasov:
@aknysh try now

MaxymVlasov:
@aknysh it will be much more helpful if you use tflint checks, which point where issues are. 
Idk why Nuru did it in this way, but I added a description to it too

image

PR #212, opened 2025-05-30
feat: Add Bottlerocket FIPS image variants

matt--williams:
## what

- Add Bottlerocket FIPS image variants

## why

- Allows users to deploy the FIPS variant, similar to the current NVIDIA variant.

## references

- None


oycyc (approved):
Looks good, confirmed with the documentation here https://docs.aws.amazon.com/eks/latest/userguide/retrieve-ami-id-bottlerocket.html

oycyc:
🙌  thanks @matt--williams

PR #214, opened 2025-06-04
EKS UserScript Bootstrapper is duplicated when using UserScript with Managed AMI

ChrisMcKee:
## what

When using the default AMI with a custom userscript AWS EKS will automatically inject the bootstrapper after your user-script.

This means 

* The userscript template will have two bootstrappers which causes the node to fail to register with the cluster
* The "after" userscript option fails as there's no way to register code after the EKS injected registration.

Before

image

After 

image

## why

* Added further information to the variable `after_cluster_joining_userdata`
* Split the bootstrapper out into its own file so it can be added only if the config allows it
* Added precondition to launch-template.tf `aws_launch_template`
* Added note to readme


ChrisMcKee:
This originally rotted out in https://github.com/cloudposse/terraform-aws-eks-node-group/pull/207

PR #216, opened 2025-06-26
Add AL2023_ARM_64_NVIDIA ami type

ricky26j:
## what

Add AL2023_ARM_64_NVIDIA ami type

## why
* This will allow users to utilize this module for these AMIs, which were first released in April


oycyc (approved):
Looks good and matches here https://docs.aws.amazon.com/eks/latest/userguide/retrieve-ami-id.html

oycyc (approved):

oycyc:
good to go @ricky26j!

PR #225, opened 2025-11-14
feat: enable node auto-repair by default

johncblandii:
## what

The node auto-repair configuration for the node group will be enabled by default.

## why

This is a best practice.

## references

DEV-3756

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -475,7 +475,7 @@ variable "node_group_terraform_timeouts" {
 
 variable "node_repair_enabled" {
   type        = bool
-  description = "The node auto-repair configuration for the node group will be enabled. Defaults to false"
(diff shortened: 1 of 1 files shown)


variables.tf
```diff
@@ -476,7 +476,7 @@ variable "node_group_terraform_timeouts" {
 variable "node_repair_enabled" {
   type        = bool
   description = "The node auto-repair configuration for the node group will be enabled. Defaults to false"
```
johncblandii:
```suggestion
  description = "The node auto-repair configuration for the node group will be enabled. Defaults to true"
```

aknysh (approved):

oycyc (approved):

